PR #3, opened 2016-05-05
first stab at referencing all shared ui components via n-ui

wheresrhys:
@i-like-robots @ironsidevsquincy


main.js
```diff
@@ -0,0 +1,23 @@
+import oDate from './date';
+
+import layout from 'n-layout';
```
i-like-robots: I guess in future the module above should be refactored into this one

wheresrhys: yep. any non origami thing referenced in here should get moved in here once I'v eroled n-ui out everywhere

PR #7, opened 2016-05-13
Integrate next buttons component and add deprecation warnings

i-like-robots:


wheresrhys:
Could you also kill off all the code in next-buttons and release as a major please. Otherwise :+1:

Issue #8, opened 2016-05-13
CSS entry points


i-like-robots:
Looking right now we have:

- bootstrap.scss
- critical.scss
- main-ie8.scss
- non-critical.scss

I guess there are a few q's about that:

- where is main.scss?
- should 'critical' styles be be left up to the consuming app?
- should we generate the IE stylesheet as suggested [here](https://github.com/Financial-Times/next-sass-setup/pull/111)?

i-like-robots:
The way I kind of see this working is like we had at LP which is the distinction of a 'core' and 'components'.

Core included the bootstrapping, low level styles and primitives. All components were opt-in by the consuming app. For example:

```
`
|- core
|      setup
|      layout
|      grid
|      typography
|      buttons
|      utils
|      date
|      ads
|- components
|      card
|      expander
|      forms
|      welcome-message
|      cookie-message
|      header
|      footer
```

And an app using n-ui:

```scss
@import "n-ui/core";
@import "n-ui/header/main";
@import "n-ui/footer/main";
@import "n-ui/header/card";
```

wheresrhys:
main entry point = bootstrap
possibly temporaril remove ie8 variant for sanity reasons
rename core => __ something else

wheresrhys:
I think we have a reasonably good thing now

PR #9, opened 2016-05-13
Re-organisation

i-like-robots:
As discussed #8 

Next to-do is to integrate n-layout

screen shot 2016-05-13 at 15 11 27


wheresrhys:
still not sure about 'foundation' as a name, but not set in stone I suppose.

It's worth doing before/after builds of an app and runnig through `find ./public -name '*.css' -exec wc -c {} \;` so check for accidental bloat

Otherwise :+1::+1::+1:

i-like-robots:
Yes, I'll keep a close eye on build sizes. So far as I can tell this should make no difference - when the dependency tree is flatter I'm sure we'll find a bunch to de-dupe =D

PR #11, opened 2016-05-13
features are off by default, with easy to turn on presets

wheresrhys:
@i-like-robots


main.js
```diff
@@ -1,26 +1,80 @@
...
-			flags.get('saveForLater') && clientOpts.push({relationship: 'saved', type: 'content'});
-			const myftClient = myFtClient.init(clientOpts);
-
-			layout.init(flags, opts);
-			oDate.init();
-			prompts.init();
-			myFtUi.init({ anonymous: !(/FTSession=/.test(document.cookie)) });
-			return Promise.resolve({flags}).then(cb);
+
+			if (opts.myft) {
```
i-like-robots: I wonder how we can do this so that these things are not bundled? We have the environment vars plugin in webpack which helps remove dead code but not sure if it prunes whole dependencies.

main.js
```diff
@@ -1,26 +1,80 @@
...
+		messagePrompts: true
+	}
+}
+
 module.exports = {
-	bootstrap: function (cb, opts = {}) {
+	bootstrap: function (opts = {}, cb) {
+
+		if (opts.preset) {
+			opts = Object.assign(presets[opts.preset], opts);
```
i-like-robots: So `bootstrap({ preset: 'all', myft: false }`, nice. Maybe options should always be applied though, E.G.:

```js
opts = Object.assign({}, opts.preset ? presets[opts.preset] : null, opts)
```

main.js
wheresrhys: I would argue as long as we keep the number of bootstrappable things to a sensible amount (which I think this is close-ish to), and move to having this as a shared bundle between apps, then the benefits of sharing identical code are worth the redundancy.

As part of this whole journey I plan to lazy load loads of stuff in js-setup too, and possibly much of the stuff in here could be lazy-loaded too, though I think the biggest win is in getting big bits of js shared between apps

main.js
wheresrhys: yep it already does that as it is, unless I've made a dumb mistake

main.js
i-like-robots: > having this as a shared bundle between apps, then the benefits of sharing identical code are worth the redundancy.

+1

main.js
i-like-robots: You can't do `Object.assign(undefined)` but you can do `Object.assign({}, undefined)` =]

i-like-robots: Oh and I meant maybe remove the conditional so the preset is optional

PR #14, opened 2016-05-16
Rhys/sass flags

wheresrhys:
So this would work something like

```
@import n-ui/configure
@include nUiConfigure((preset: all, welcomeMessage: false))
@import n-ui/bootstrap
```
I forget exactly why I thought this would be a good idea. I think mainly because it looks more like the js api. Also allows a clearer separation between wanting a feature to be available, and silent mode flagging (which is more of an implementation detail that we may not want consumers to touch directly)


configure.scss
```diff
@@ -0,0 +1,53 @@
+$_n-ui-features: (
+	ads: false,
+	buttons: false,
+	grid: false,
+	colors: false,
+	icons: false,
+	typography: false,
+	cookieMessage: false,
```
i-like-robots: I guess we should distinguish between 'foundations' (starting to dislike that already) and the non-core, opt-in 'components'. Have tried to define them a little in [the wiki](https://github.com/Financial-Times/n-ui/wiki).

i-like-robots: Oh yeah, and everything in Sass is lower case hyphenated - the only time I've seen camelCase in Sass is Origami. I wonder if we should pick a side.

wheresrhys: I'm on the fence as to whether the foundations/components distinction is worth having. Adds verbosity and, if we want to preserve the distinction at all levels of the API, complexity and demands on the user's memory.

i-like-robots: The reason I quite like sorting modules into folders is that there's no need to lookup and open files, it only takes a glance to see a directory structure - either locally or on Github.

I think the ability to quickly lookup what's bundled by default and what isn't is very useful but not wedded to any particular way of doing that.

i-like-robots:
Just so it's recorded somewhere from our offline chat:

- expect configuration _before_ bootstrap is imported, throw `@error` if it is not
- output a `/*! loud */` comment displaying configuration info
- primitives (E.G. buttons, colors, icons) are not configurable

wheresrhys:
... and in addition to the last point, if possible we should try and output a sensible subset of e.g. o-buttons, o-grid if origami allows

README.md
```diff
@@ -1,50 +1,38 @@
...
+	if (flags.get('konami')) {
+		easterEgg.init();
+	}
+})
+```
 
-### Reduce duplication
+```Sass
+@import 'n-ui/configure';
+@include nUiConfigure((
```
i-like-robots: Should this be a function rather than a mixin as it has no output? Maybe print the config here?

configure.scss
```diff
@@ -0,0 +1,68 @@
...
+	message-prompts: false,
```
i-like-robots: mix of camelCase and hyphenated keys

README.md
wheresrhys: good call

configure.scss
wheresrhys: Ah that's a mistake. I think we shoudl go camel case

configure.scss
```diff
@@ -0,0 +1,70 @@
...
+
+	@if map-has-key($opts, preset) {
+		$preset: map-get($_n-ui-presets, map-get($opts, preset));
+		$_n-ui-features: map-merge($_n-ui-features, $preset) !global;
+		$opts: map-remove($opts, preset);
+	}
+
+	$_n-ui-features: map-merge($_n-ui-features, $opts) !global;
+
+	/* n-ui features: */
```
i-like-robots: Could make this a one liner maybe:

```scss
/*! n-ui features: #{inspect($_n-ui-features)} */
```
```

PR #18, opened 2016-05-17
lightweight import of o-grid

wheresrhys:
@i-like-robots

### DIFF
--- a/grid/main.scss
+++ b/grid/main.scss
@@ -1,6 +1,104 @@
-@if nUiHas(grid) {
-	$o-grid-is-silent: false;
-}
+$_n-ui-grid-applied: false !default;
+
+// Set gutters to be slightly smaller than the default on larger screens
+$o-grid-gutters: (
+	default: 10px,
+	M:       16px,
+);
 
 @import 'o-grid/main';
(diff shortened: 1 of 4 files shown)


grid/main.scss
```diff
@@ -1,6 +1,104 @@
...
+
+					// Apply width in %
+					@include _oGridFixSafariWrap($colspan);
+					min-width: oGridColspan($colspan);
+					max-width: oGridColspan($colspan);
+					width: oGridColspan($colspan);
+				}
+			}
+		}
+	}
```
i-like-robots: As discussed elsewhere, I don't like this as it creates so much cruft... I can only find a few uses of it within o- components but I wonder if we could add a warning here saying you should generate a separate stylesheet or something?

wheresrhys: don't like what exactly?

i-like-robots: Usage of this mixin. IMO we should where possible encourage generating a separate stylesheet for old IE at a fixed breakpoint.

wheresrhys: let's revisit before releasing v1.

PR #19, opened 2016-05-17
copied icons sass from nex-sass-setup to n-ui

wheresrhys:
@i-like-robots

Waiting on a minor release of o-icons before it actually works (already merged into master)


i-like-robots:
LGTM

PR #20, opened 2016-05-17
Small fixes to grammar, formatting, critical pragmas and paths

i-like-robots:


configure.scss
```diff
@@ -1,20 +1,20 @@
 $_n-ui-features: (
 	ads: true,
 	buttons: true,
-	grid: true,
-	// always in silent mode
-	colors: false,
-	// always in silent mode
-	icons: false,
-	typography: true,
+	card: false,
```
wheresrhys: colors & icons

Even though config here won't do anything I think this list is a good as a living list of what gets included

configure.scss
```diff
@@ -1,20 +1,20 @@
...
 	myft: false,
 	notification: false,
-	welcomeMessage: false,
-	header: true,
-	footer: true
-);
+	overlay: false,
+	typography: true,
+	welcomeMessage: false
+) !default;
```
wheresrhys: why !default?

i-like-robots: Just in case this is accidentally included again which would mean this assignment would overwrite the configured state.

We could also throw a wobbly.

configure.scss
i-like-robots: have added all modules with null values

configure.scss
i-like-robots: wobbly now thrown

PR #35, opened 2016-05-19
disabling human readable mode in grid

wheresrhys:


grid/README.md
```diff
@@ -23,4 +21,4 @@ e.g.  `@include nUiGridOffset(M, 3, pull)`
 
 ## Human readable selectors
 
-Pending a pull request to o-grid, these will be disabled, so use the numeric selectors i.e. `data-o-grid-colspan="6"` not `data-o-grid-colspan="half"`. The one exception is the `hide` keyword
+Pending a pull request to o-grid, these will be disabled, so use the numeric selectors i.e. `data-o-grid-colspan="6"` not `data-o-grid-colspan="half"`. Notably, use `0` instead of `hide`.
```
i-like-robots: I think this will break things =[

wheresrhys: Why? `0` doesn't actually work, but should (just an alias for `hide`)

Needs a PR in o-grid to uncouple from the other human readable selectors though

wheresrhys: Either way, not planning on merging til that's fixed upstream

i-like-robots: I mean I believe we use `hide` a fair bit in our templates

PR #37, opened 2016-05-20
added mixin for critical start/end

wheresrhys:


_sass-utils/_main.scss
```diff
@@ -1,3 +1,38 @@
...
+		@return '! end:head.css ';
+	}
+
+	@return '';
+}
+
+
+@mixin nUiCriticalStart() {
+	@if nUiHas('critical') {
+		/*! start:head.css*/
```
ironsidevsquincy: couldn't these call the functions?

ironsidevsquincy: Oh, I see, they're being deprecated

PR #38, opened 2016-05-23
Add class to document when new header flag is on

i-like-robots:
So we don't have to add it with JS and cause a hefty reflow =]


wheresrhys:
d'you want to also move to v1.0.0 in this? @ironsidevsquincy See any reason not to?

ironsidevsquincy:
Might as well, being used everywhere now

i-like-robots:
💥

Issue #41, opened 2016-05-23
Deprecte nextIcon


wheresrhys:
WIP splitting of nextIcon to handle brands differently. It'll probably mean we have to go through deleting all calsl to nextIcon TBH

```
$_n-ui-brands-list: brand-ft-masthead, myft-bolder, brand-ft, brand-fast;

@mixin nUiBrand($brand-name, $size: 20, $color: null) {
	display: inline-block;
	width: $size + 0px;
	height: $size + 0px;

	$logo-url: 'https://next-geebee.ft.com/image/v1/images/raw/ftlogo:#{$brand-name}';
	$query: '?source=next';

	@if $color != null {
		$color: toHex($color);
		$query: $query + '&tint=%23#{$color},%23#{$color}';
	}

	background-image: url($logo-url + $query + '&format=svg');
	// ie7/8 fallback
	background-image: url($logo-url + $query + '&format=png&width=#{$size}')\9;

	background-repeat: no-repeat;
	background-size: contain;
	background-position: 50%;
	background-color: transparent;
}

@mixin nextIcon($icon-name, $color: null, $size: 20, $icon-only: false) {
	@warn '`nextIcon` is deprecated, use `oIconsGetIcon` or `nUiBrand` instead';

	@if (index($_n-ui-brands-list, $icon-name)) {
		@include nUiBrand($icon-name, $size, $color);
	} @else {
		$apply-base-styles: $icon-only == false;
		@include oIconsGetIcon($icon-name, $color, $size, $size, $apply-base-styles);
	}
}
```

PR #42, opened 2016-05-24
Rhys/next icon brands

wheresrhys:


wheresrhys:

The new mixin I'd written https://github.com/Financial-Times/n-ui/issues/41 didn't actually work properly, but we were saved by sass setup being imported all over the place, and overwriting nextIcon in the process. Best to look at again once sass-setup is properly gone

Issue #44, opened 2016-05-24
next-welcome display bug


tavvy:
Firefox Version 46.01

<img width="1392" alt="screen shot 2016-05-26 at 09 38 40" src="https://cloud.githubusercontent.com/assets/660635/15568790/afbfeebc-2325-11e6-9026-bdb7b7509285.png">

PR #48, opened 2016-05-25
proposed fix for #44

tavvy:


ironsidevsquincy:
Does that mean we don't need https://github.com/Financial-Times/n-ui/blob/master/welcome-message/index.js#L28, as it now already starts with that class?

tavvy:
Im confused by what the logic is suppose to be here.
The `fixbar` is a simplified version (without copyright links etc) of the default.

IF "`// Don't display the welcome bar if already acknowledged, or if we can't store in storage`" then return without displaying the `fixbar`.

I am observing that when we do display the fix bar, the close button will `unfixBar` which hides the `fixbar` and displays the default? But they are seemingly the same thing.

ironsidevsquincy:
They are slightly different, but yes, that's right. Show the fixed (sticky) bar first, and if the user closes show the bar at the bottom of the screen always

tavvy:
Right I see. Well I think I have inadvertently fixed a different bug to the one I was originally trying to address.

And actually looking at what this tries to achieve it doesn't change how the component behaves with js disabled anyway. So I'm going to close this and go back to the drawing board.

PR #50, opened 2016-05-25
added env.scss entry point 

wheresrhys:
(to replace sass-setup for components not imported yet)

@i-like-robots @ironsidevsquincy


ironsidevsquincy:
👍

Issue #54, opened 2016-05-26
next-welcome & discrete preset


tavvy:
The next-welcome component can and will render despite using the discrete preset (or setting `welcomeMessage: false`). I believe this is because of this logic: https://github.com/Financial-Times/n-ui/blob/master/layout/wrapper.html#L66 that allows the template to be pulled in if the `flag` is present.

Now the solution would be for the welcome message to render in a hidden state, and only be displayed if the scss and js are pulled in and the component is init'd.

The solution couldn't involve changes to the components scss or js as if running in discrete modes these wouldn't actually be pulled in.

PR #55, opened 2016-05-26
Fix welcome message in discrete mode #54

tavvy:
see https://github.com/Financial-Times/n-ui/issues/54

### DIFF
--- a/welcome-message/main.scss
+++ b/welcome-message/main.scss
@@ -7,6 +7,12 @@
 	border-top: 5px solid getColor('grey-tint3');
 	position: relative;
 
+	/* resets .n-util-hide which is used in template.html
+	 * when this feature is disabled (discrete preset) it will be hidden
+	 * when enabled this scss will be pulled in and the component will render
+	 */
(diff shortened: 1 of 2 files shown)


ironsidevsquincy:
I wouldn't use the `n-util-hide` class, as the implementation might change (so `display: block` doesn't unhide). Also, you'll need to wrap `n-welcome`'s css in a conditional, e.g. https://github.com/Financial-Times/n-ui/blob/master/header/main.scss

So something like

```
$_n-ui-welcome-message-applied: false;

.n-welcome {
    display: none;
}

@if $_n-ui-welcome-message-applied == false and nUiHas(welcomeMessage) {
    .n-welcome {
        ...
        display: block;
        ...
    }

    $_n-ui-welcome-message-applied: true !global;
}
```

tavvy:
Great, that is far neater!
https://github.com/Financial-Times/n-ui/pull/55/files?w=1

ironsidevsquincy:
👍

PR #58, opened 2016-05-27
Defer configuring until after polyfill has loaded

ironsidevsquincy:
As some browsers need `Object.assign` polyfilled

@wheresrhys


wheresrhys:
:+1:

PR #64, opened 2016-05-31
Integrates n-messages

ifyio:
cc/ @i-like-robots @wheresrhys


ifyio:
removing the .idea folder

configure.scss
```diff
@@ -11,6 +11,7 @@ $_n-ui-features: (
 	grid: true,
 	header: true,
 	icons: null,
+	messages: true,
```
i-like-robots: should be false I reckon

i-like-robots:
Looks fine to me - is the plan to copy over the messages code into this repo?

configure.scss
wheresrhys: should be `false` here, otherwise they would get initialised on error pages, opt out form etc

wheresrhys:
Looks good (aside from 1 comment). 

Will this in general only be used for promotional messages? If it will, could it be renamed to promo-messages or something along those lines, to make clearer the distinction between it and notifications.

ifyio:
@i-like-robots Yes it is. Eventually.

ifyio:
@wheresrhys yes that's a good idea. Will do that now

PR #80, opened 2016-06-06
make colspan extend for each layout optional

laurieboyes:
to cater for apps using a fixed o-grid layout, which may omit some

@wheresrhys


wheresrhys:
:+1:

PR #89, opened 2016-06-07
Load gpt and ads-api in parallel

adgad:
@VladDubrovskis @andrewgeorgiou1981 @leggsimon @wheresrhys 

A bit hacky, but call init(withoutTargeting), and then call oAds.config(withAllTargeting) before initSlot is called. Should speed things up a wee bit.

Previously:
[ads-api/user, ads-api/content] -> GPT calls -> Ad calls

Now: 
[ads-api/user, ads-api/content, GPT calls] -> Ad calls

### DIFF
--- a/ads/index.js
+++ b/ads/index.js
@@ -48,19 +48,21 @@ function getUserTargetingPromise () {
 		.catch(() => ({}));
 };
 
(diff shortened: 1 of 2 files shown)


leggsimon:
LGTM

PR #92, opened 2016-06-08
display flex helper

wheresrhys:
nicked from myft-page @i-like-robots

### DIFF
--- a/_sass-utils/_mixins.scss
+++ b/_sass-utils/_mixins.scss
@@ -100,3 +100,19 @@
 	@warn '`nCrosshatchBG` has been deprecated, use `nUiPinstripeBG` instead';
 	@include nUiPinstripeBG($args...);
 }
+
+/// Blackberry & old iOS safe display: flex declaration
+/// @output 'display: flex', with only the vendor prefixes we trust
+@mixin nDisplayFlex() {
+	display: block;
+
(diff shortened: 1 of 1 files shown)


i-like-robots:
I wonder if this can be dealt with by autoprefixer?

i-like-robots:
Just wondering if we'd also want to apply this to the grid and any other use of `display: flex`?

wheresrhys:
~~Nope. It took a lot of hackery to do this. Working around autoprefixer was half the problem. It does all the prefixes https://github.com/postcss/autoprefixer#flexbox-filters-etc~~

It looks like they have indeed fixed it! added a `flexbox: "no-2009"` option https://github.com/postcss/autoprefixer/commit/b6d228fe4b5fc6f326d5bc80d0300ca157147db7, https://github.com/postcss/autoprefixer#disabling. This will need careful coordination with origami cc @aelias

i-like-robots:
So this would make the following non-flexi:

- Android 4.3
- BB7
- iOS6

That's probably OK...

wheresrhys:
We deliberately suppress flexbox (where we remember to) in those browsers as it is totally broken. Twas the cause of all the pain in taking over m.ft.com traffic in december if you recall

wheresrhys:
closed in favour of https://github.com/Financial-Times/n-makefile/pull/86

PR #94, opened 2016-06-09
Track first ad loaded

leggsimon:
Adding an oTracking event to gather `firstAdLoaded` performance mark. The `firstAdLoaded` mark was previously being very rarely sent using `n-instrumentation` due to the `onLoad` event happening before the first ad was being loaded.

cc/ @adgad @commuterjoy @ironsidevsquincy @wheresrhys

### DIFF
--- a/ads/index.js
+++ b/ads/index.js
@@ -74,7 +74,7 @@ function onAdsComplete (event) {
 			utils.log.info('Ad loaded in slot', event);
 			if (slotsRendered === 0) {
 				perfMark('firstAdLoaded');
-				const firstAdLoaded = performance.getEntriesByType ?
+				const marks = performance.getEntriesByType ?
 					performance.getEntriesByType('mark')
 						.filter(mark => mark.name === 'firstAdLoaded')
 						.reduce((marks, mark) => {
@@ -86,7 +86,7 @@ function onAdsComplete (event) {
 				broadcast('oTracking.event', {
 					category: 'page-load',
 					action: 'timing',
-					timings: { firstAdLoaded }
+					timings: { marks }
 				});
 			}
 		} else if (detail.slot.gpt && detail.slot.gpt.isEmpty === true) {
--- a/ads/index.js
+++ b/ads/index.js
@@ -7,6 +7,8 @@ const jsonpFetch = require('n-jsonp');
 
 import { perfMark } from '../utils'
 import { broadcast } from '../utils'
+import { getMarks } from 'n-instrumentation';
+
 
 let slotCount;
 let slotsRendered = 0;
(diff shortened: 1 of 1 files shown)


ads/index.js
```diff
@@ -73,6 +74,9 @@ function onAdsComplete (event) {
 			utils.log.info('Ad loaded in slot', event);
 			if (slotsRendered === 0) {
 				perfMark('firstAdLoaded');
+				if (!/spoor-id=0/.test(document.cookie)) {
```
adgad: Should we use a different number here? I have a feeling the poor people with spoor-ids starting with 0 are getting a raw deal with all our sampling.

ads/index.js
```diff
@@ -87,6 +91,24 @@ function onAdsComplete (event) {
 	}
 }
 
+function sendAdLoadedTrackingEvent() {
+	const marks = performance.getEntriesByType ?
+		performance.getEntriesByType('mark')
```
adgad: I think you can also do `getEntriesByName` here, and saving having to look through the array

leggsimon:
💭  Just thinking, I guess we'll now need to ensure that n-instrumentation doesn't send the `firstAdLoaded` performance mark if it does by any chance have it before it sends them all to beacon.

wheresrhys:
why have a separate event rather than delay the main navigation timing event?

ads/index.js
adgad: Maybe we could also add an offset from domContentLoaded?

ads/index.js
leggsimon: wouldn't we want all our performance marks to come from the same sample though?

leggsimon:
@wheresrhys as @ironsidevsquincy said yesterday, you either risk not having the data if someone navigates away from the page before 5000ms (for example) or you bias your data towards users that stay on a page for 5000ms

ads/index.js
adgad: I don't think it really matters, since we can't tie the two events up nicely anyway

wheresrhys:
I'd say the risk is overstated, but don't feel that strongly

ads/index.js
```diff
@@ -87,6 +91,23 @@ function onAdsComplete (event) {
...
+	const marks = performance.getEntriesByType ?
+		performance.getEntriesByName('firstAdLoaded')
+			.reduce((marks, mark) => {
+				marks[mark.name] = Math.round(mark.startTime);
+				return marks;
+			}, {}) :
+		{};
+
+	broadcast('oTracking.event', {
+		category: 'page-load',
```
commuterjoy: `category: 'ads', action: 'first-load'` is probably more obvious, as you already have `ads:blocked`

adgad:
:+1

ads/index.js
leggsimon: Turns out there's a mistake here and actually the "poor people with spoor-ids starting with 0" are specifiacally *not* getting targeted. Fix [here](https://github.com/Financial-Times/n-ui/pull/104)

PR #103, opened 2016-06-13
added nUiCritical mixin

wheresrhys:
IMHO

````
@include nUiCritical() {
   // some styles
}
```

Is easier to comprehend

````
@include nUiCriticalStart;
   // some styles
@include nUiCriticalEnd;
```

provided `//some styles` doe snot use an `@import`

@i-like-robots 

Although I am wary of increasing the API surface area


i-like-robots:
LGTM

PR #112, opened 2016-06-14
Make ad loading faster (behind flag)

adgad:
/cc @wheresrhys @ironsidevsquincy @leggsimon 
The faster ads load, the better viewability/potentially impressions.

By not waiting for ftNextLoaded, ads should load ~4-500ms faster. Do this behind a flag so we can set up a speedcurve page and check the impact on other metrics.

### DIFF
--- a/ads/index.js
+++ b/ads/index.js
@@ -162,8 +162,12 @@ module.exports = {
 		})
 	},
 	init: function (flags) {
-		window.addEventListener('ftNextLoaded', () => {
+		if(flags && flags.get('adsPrioritiseLoading')) {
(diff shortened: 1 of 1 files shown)


wheresrhys:
Are these metrics coming through to beacon? Speedcurve isn't really the best for actually making decisions IMHO as it's far too wobbly (though it is good for communicating about perf to others).

:+1:

adgad:
We have the most important one coming through in beacon (https://beacon.ft.com/dashboard/advertising/). Although we'd have to turn on flag/AB test to check that.

PR #121, opened 2016-06-15
allow multiple calls to bootstrap so nav etc can be initialised super quick

wheresrhys:
@ironsidevsquincy


_deploy/wrapper.js
```diff
@@ -1,5 +1,13 @@
 /*global ftNextFireCondition*/
-module.exports = require('n-ui');
+const nUi = require('n-ui');
+
+nUi.configure({
+	preset: 'discrete'
```
wheresrhys: read the conf off an env var. default to discrete

PR #148, opened 2016-06-24
Footer

aintgoin2goa:


footer/_footer.html
```diff
@@ -0,0 +1,44 @@
+<footer class="o-footer" data-o-component="o-footer" data-trackable="footer">
```
i-like-robots: `role="contentinfo"`

footer/_footer.html
```diff
@@ -0,0 +1,44 @@
...
+				<span>&#xA9; THE FINANCIAL TIMES LTD 2016.</span>
+				<span><abbr title="Financial Times">FT</abbr> and &#x2018;Financial Times&#x2019; are trademarks of The Financial Times Ltd.</span>
+			</div>
+		</div>
+	</div>
+	<div class="o-footer__brand">
+		<div class="o-footer__container">
+			<div class="o-footer__row">
+				<div class="o-footer__col o-footer__col--full-width">
+					<div class="o-footer__brand-logo" aria-label="A Nikkei Company"></div>
```
i-like-robots: no use for the aria label here, there's nothing to focus on - maybe some visually hidden text instead?

footer/partials/_footer_links.html
```diff
@@ -0,0 +1,11 @@
+
+<nav class="o-footer__nav" data-trackable="{{name}}">
+	<h3 class="o-footer__title">{{title}}</h3>
```
i-like-robots: Heading level should probably be below any subtitles in the content - h5?

layout/wrapper.html
```diff
@@ -66,7 +66,8 @@
 				{{#if @root.flags.welcomePanel}}
 					{{>n-ui/welcome-message/template}}
 				{{/if}}
-				{{>n-ui/footer/template content=footer}}
+				{{>n-ui/footer/template}}
+				{{>n-ui/drawer/template content=footer}}
```
i-like-robots: no need for the `content=footer`

main.js
```diff
@@ -85,6 +88,10 @@ export function bootstrap (cb) {
 			header.init(flags);
 		}
 
+		if(opts.footer){
```
i-like-robots: linting =]

viewport/index.js
```diff
@@ -1 +1,45 @@
...
+	if(listening){
+		return;
+	}
+	let lastBreakpoint = getCurrentBreakpoint();
+	fireBreakpointEvent(lastBreakpoint);
+	oViewport.listenTo('resize');
+	document.body.addEventListener('oViewport.resize', () => {
+		let breakpoint = getCurrentBreakpoint();
+		if(breakpoint !== lastBreakpoint){
+			fireBreakpointEvent(breakpoint);
```
i-like-robots: This should definitely be a feature of o-grid I reckon!

aintgoin2goa: oops - some of it already is (just not the event).  

Do you think this code should go in `/grid` rather than `/viewport`  ?

i-like-robots: grid has the breakpoints I guess

viewport/index.js
```diff
@@ -1 +1,45 @@
...
+const oViewport = require('o-viewport');
+const utils = require('../utils');
+
+const GRID_BREAKPOINTS = new Map([
+	['default', [0,489]],
+	['S', [490,739]],
+	['M', [740,979]],
+	['L', [980,1219]],
+	['XL', [1220,Infinity]]
+]);
```
i-like-robots: I just realised o-grid (which should be bundled) has a `getCurrentLayout` method so we probably don't need to dupe the breakpoints

PR #149, opened 2016-06-24
Add new myFT UI component for grouped preference buttons

keirog:
It doesn't look amazing, but hopefully good enough while we build the new prefs page.

Mini screens:
mini

Big screens:
big

Still to come: JS changes to get the new preference type working for enhanced users.

### DIFF
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -71,6 +71,53 @@ $_n-ui-myft-is-silent: nUiHas('myft') == false or $_n-ui-myft-applied == true;
 		}
 	}
 
+	.n-myft-ui--prefer-group {
+		@include oGridRespondTo($from: 'M') {
+			padding-left: $spacing-unit;
+		}
+		width: 100%;
+
+		.n-myft-ui__button {
+			border-radius: 0;
+			margin-left: -5px;
+		}
+		.n-myft-ui__button:first-of-type {
+			border-bottom-left-radius: 3px;
+			border-top-left-radius: 3px;
+			border-left-width: 1px;
+		}
+
+		.n-myft-ui__button:last-child {
+			border-bottom-right-radius: 3px;
+			border-top-right-radius: 3px;
+		}
+
+		.n-myft-ui__button[aria-pressed="true"] {
+			padding-left: 26px;
+			position: relative;
+		}
+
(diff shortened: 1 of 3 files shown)


myft/main.scss
```diff
@@ -71,6 +71,53 @@ $_n-ui-myft-is-silent: nUiHas('myft') == false or $_n-ui-myft-applied == true;
 		}
 	}
 
+	.n-myft-ui--prefer-group {
+		@include oGridRespondTo($from: 'M') {
+			padding-left: $spacing-unit;
+		}
```
ironsidevsquincy: does sass lint get funny about this rule ordering?

keirog: It didn't throw an error; should it? I was thinking the ordering was something like:
1. `include`s with no code blocks
2. `include`s with code blocks
3. other stuff

keirog: Ah, it's just a warning but I'll amend the ordering. It wants this:
`@extends`, `@includes` without `@content`, properties, `@includes` with `@content`

ironsidevsquincy: Unrelated to this, but maybe we should think about, in general, upgrading all our sass warnings to errors? Think the next codebase as a whole is in a better place than it was

keirog: Am definitely up for that

myft/templates/prefer-group.html
```diff
@@ -0,0 +1,24 @@
...
+			data-trackable="set-{{../name}}-{{name}}"
+			name="_rel.selected" value="{{name}}"
+		>{{text}}</button>
+	{{/each}}
+		<button
+			type="submit"
+			class="n-myft-ui__button" aria-pressed="{{#if selected}}false{{else}}true{{/if}}"
+			data-trackable="set-{{name}}-off"
+			name="method" value="delete"
+		>Off</button>
```
phamann: Why is off not passed in as an option? What if I wanted a group without off?

phamann:
👍 LGTM other than comments

myft/templates/prefer-group.html
keirog: I was thinking that any non-binary preferences would still always need an off state, but I guess not. Happy to amend

keirog:
(Will amend to make "off" live in the viewModel later. Keen to get this in prod)

PR #161, opened 2016-06-27
return script in ftNextLoadScript

lc512k:
@wheresrhys @ironsidevsquincy 

`ftNextLoadScript` was not returning anything, so we were getting an undefined error in `loadScript` in https://github.com/Financial-Times/n-ui/blob/master/utils/index.js#L35


callumlocke:
LGTM 👍

PR #166, opened 2016-06-28
Add enhanced functionality for grouped preferences

keirog:
Turn on the `myFtDigestPrefsEnhanced` flag to see changes on the alerts page/product tour.

I *really* want to refactor this whole component…

prefs
(I clicked a normal on/off toggle in the gif just to prove that still worked)

### DIFF
--- a/myft/ui-browser.js
+++ b/myft/ui-browser.js
@@ -341,7 +341,7 @@ function getInteractionHandler (myftFeature) {
 	return function (ev, el) {
 		ev.preventDefault();
 
-		const buttonWithValTriggered = !!(el.tagName.toLowerCase() === 'button' && el.name && el.value);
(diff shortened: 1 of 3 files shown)


myft/ui-browser.js
```diff
@@ -310,43 +314,56 @@ function onLoad (ev) {
...
 		state: true
 	});
 }
 
+// extract properties with _rel. prefix into nested object, as expected by the API for relationship props
+function extractMetaData(inputs) {
+	const meta = {};
+
+	inputs.forEach((input) => {
+		if (input.name.startsWith('_rel.')) {
```
phamann: is `startsWith` safe to use?

phamann:
👍 LGTM other than comments

myft/ui-browser.js
keirog: Yep, polyfilled

PR #171, opened 2016-06-28
Update follow-email to work with the new digest

keirog:
Updates the follow+email overlay to reflect changes to the email digest preference model.

Also sneaks in the missing timezone data when setting the preference (which is just hard-coded for now, like the prefs page).


myft/js/set-follow-preferences.js
```diff
@@ -1,11 +1,15 @@
 const myFtApi = require('next-myft-client');
+const relData = {
+	timezone: 'Europe/London',
```
keirog: Needs changing when we give users the choice, obvs

PR #176, opened 2016-06-29
some protips

wheresrhys:
Added a protips doc

@i-like-robots  @ironsidevsquincy 

An experiment to output a pro tip every time n-webpack runs. Might be an effective way of disseminating information about how to us n-ui?


wheresrhys:
see also https://github.com/Financial-Times/n-webpack/pull/11

PR #182, opened 2016-06-30
Move ad reporting from o-ads to n-ui

adgad:
/cc @tavvy @VladDubrovskis @leggsimon 

Moves the code from https://github.com/Financial-Times/o-ads/pull/52 to here, since I don't think it's ready for the prime time o-ads hotspot yet.
Updates styling a bit so it's more visible over ads, and update the link when slot refreshes.


tavvy:
👍

PR #186, opened 2016-06-30
React save template

bjfletcher:
@ironsidevsquincy There's the existing HTML template which I've ported to become a React component. I've gone with a different boilerplate where properties are defined in the arguments rather than in comments. I hope this is OK? I prefer this approach.

### DIFF
--- a/myft/templates/save-for-later.js
+++ b/myft/templates/save-for-later.js
@@ -1,7 +1,14 @@
 import React from 'react';
 
-export default ({ myFtApiWrite, contentId, variant, buttonText }) => {
-	return !myFtApiWrite ? null : (
+/**
+ * @param {string} contentId
+ * @param {boolean} [myFtApiWrite = false]
+ * @param {string} [variant]
+ * @param {string} [buttonText]
+ */
+
+export default ({ contentId, myFtApiWrite, variant, buttonText }) => {
+	return myFtApiWrite ? (
 		<form className="n-myft-ui n-myft-ui--save" method="POST"
 				data-content-id={contentId}
 				data-myft-ui="saved"
@@ -17,5 +24,5 @@ export default ({ myFtApiWrite, contentId, variant, buttonText }) => {
 					title="Save to myFT"
 			>{buttonText || 'Save'}</button>
 		</form>
-	)
+	) : null;
 }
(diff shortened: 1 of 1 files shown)


ironsidevsquincy:
It would be good to have a comment above to explain what the arguments mean/what type they are (e.g. `myFtApiWrite`?)

myft/templates/save-for-later.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+export default ({ myFtApiWrite, contentId, variant, buttonText }) => {
```
ironsidevsquincy: in general, arguments in an object are optional. looks like `contentId` is a required argument, so would right this as

```
(contentId, { myFtApiWrite, variant, buttonText })
```

myft/templates/save-for-later.js
```diff
@@ -0,0 +1,21 @@
...
+	return !myFtApiWrite ? null : (
```
ironsidevsquincy: isn't of negating, why not

```
myFtApiWrite ?
    <form ...>
     ...
   </form> : 
   null;
```

myft/templates/save-for-later.js
```diff
@@ -0,0 +1,21 @@
...
+					type="submit"
+					className={`n-myft-ui__button${variant ? 'n-myft-ui__button--' + variant : ''}`}
+					data-trackable="save-for-later"
+					data-alternate-label="Saved to myFT"
+					data-alternate-text="Saved"
+					data-content-id={contentId}
+					aria-label="Save to myFT"
+					title="Save to myFT"
+			>{buttonText || 'Save'}</button>
+		</form>
```
ironsidevsquincy: is this indentation a bit off?

myft/templates/save-for-later.js
bjfletcher: For readability. I think it's better to have the very large block at the end, rather in the middle; if you want to know what happens if it's false, you'd have to scroll all the way down to see that it's `null`. What do you think?

myft/templates/save-for-later.js
bjfletcher: I tend to like have 2 tabs for continuing lines, 1 tab for actual indention. If we have 1 tab for continuing lines, then it's hard to distinguish between continuing lines and actual indention. Interested to know your opinion on this.

myft/templates/save-for-later.js
bjfletcher: That's great. :+1: :)

myft/templates/save-for-later.js
ironsidevsquincy: Looks funny to my eye, but don't have a strong opinion about it

myft/templates/save-for-later.js
ironsidevsquincy: Always find negations harder to understand, but that's just me, don't have a string opinion about it

PR #187, opened 2016-06-30
set custom js exists flag

wheresrhys:


wheresrhys:
woudl only be useful for a couple of apps at present. and feels messy

PR #189, opened 2016-06-30
const in IE < 11 = 💥

keirog:


wheresrhys:
hang on, this should be getting babelified. There's something wrong somewhere else

wheresrhys:
prop need to add an `include` to the webpack options

keirog:
Oops sorry, jumped the gun a bit 😬

PR #195, opened 2016-07-04
Pull out useragent stuff into separate module; fix iOS regex…

matthew-andrews:
…so that it only offers web app button to supported User Agents.


aintgoin2goa:
👍

PR #196, opened 2016-07-04
Convert Android detection logic to regex

matthew-andrews:


aintgoin2goa:
Ok, so the tests pass so it must work but I don't understand how that regex can match Android > 4??

aintgoin2goa:
Never mind, just worked it out.  We should be all good until Android 10 comes out!
👍

matthew-andrews:
should be a while :)

PR #198, opened 2016-07-04
Add preconnect for GPT libraries, and refactor ads init

adgad:
/cc @wheresrhys @ironsidevsquincy @andrewgeorgiou1981 @VladDubrovskis 
* Deflagify the ads init change (to not wait for ftNextLoaded)
* Add a bunch of preconnect headers for Ad domains (behind flag, so that we can check the impact on webpagetest).


andrewgeorgiou1981:
👍

wheresrhys:
Could have a 3rd variant too - `rel="dns-preconnect"`

adgad:
Not a bad idea. I think preconnect should perform better in WPT (because of the TLS) - but dns-prefetch has better browser support so would probably have wider real world impact.

PR #201, opened 2016-07-05
Rhys/non min flag

wheresrhys:
@matthew-andrews

### DIFF
--- a/_deploy/s3.js
+++ b/_deploy/s3.js
@@ -10,7 +10,7 @@ if (!versions[0]) {
 }
 
 Promise.all(versions.map(version => {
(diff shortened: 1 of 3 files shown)


wheresrhys:
In addition to this jon nangle is adding a rule to akamai to add a `no-cache` rule, which will force etag/last-modified to be used once max-age has expired (I had this rule initially, but couldn't remember why, and Rik and me agreed to remove it in order to get the S3 header to propagate to the edge. All this will return to being done in code once we move to fastly for geebee)

I will 'roll' this out cautiously i.e. get loads of people to turn on the flag, and then go for a release some morning this week

PR #204, opened 2016-07-06
Wait for Krux script to load before firing tags

adgad:
/cc @VladDubrovskis 

Wait for Krux script to load before firing tags


VladDubrovskis:
👍

PR #207, opened 2016-07-07
Mouseflow for the 1%

lc512k:
and `/signup`

cc @commuterjoy @adambraimbridge

### DIFF
--- a/tracking/js/third-party/mouseflow.js
+++ b/tracking/js/third-party/mouseflow.js
@@ -6,6 +6,12 @@ const getCookieValue = function (key) {
 	return (a) ? a[1] : undefined;
 }
 
+function enableMouseflow () {
+	window._mfq = window._mfq || [];
+	window._mfq.push(['setVariable', 'spoorId', getCookieValue('spoor-id')]);
+	loadScript('https://cdn.mouseflow.com/projects/3d6fc486-2914-4efc-a5ae-35a5eac972f2.js');
+}
+
 // Loads mouseflow tracking code
 module.exports = function (flags) {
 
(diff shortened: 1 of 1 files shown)


tracking/js/third-party/mouseflow.js
```diff
@@ -6,14 +6,44 @@ const getCookieValue = function (key) {
...
+					case 404:
+						return {};
+					case 200:
+						return response.json();
+					default:
+						throw new Error(`${response.status} - ${response.statusTest}`);
+				}
+			})
+			.then (function (session) {
+				let isNavigationSample = parseInt(session.passportId) % 100 === 0; // 1%
```
adambraimbridge: Perhaps `device.spoorId` instead of `session.passportId `?

lc512k: `passportId` seemed like a nice, evenly distributed, digit-only string I could quickly `parseInt` & `mod` :) Also @commuterjoy mentioned spoorIds not having an even distribution IIRC?

lc512k:
@commuterjoy any (`parseInt`able, evenly distrd) alternatives to passportId if that's being deprecated?

tracking/js/third-party/mouseflow.js
```diff
@@ -6,14 +6,45 @@ const getCookieValue = function (key) {
...
+					case 404:
+						return {};
+					case 200:
+						return response.json();
+					default:
+						throw new Error(`${response.status} - ${response.statusTest}`);
+				}
+			})
+			.then (function (session) {
+				if (parseInt(session.passportId) % 100 === 0) { // 1%
```
commuterjoy: this is deprecated, so perhaps switch over to the uuid at some point?

lc512k: @commuterjoy how is uuid generated? Can I safely `mod 100` over the decimal value of the last segment, ie `86976528967924` for `4f1acd1ce8f4` of `c58d7528-d4d5-4cd3-aff8-4f1acd1ce8f4`?

PR #213, opened 2016-07-08
import buttons after icons to force geebee

wheresrhys:
otherwise 
image

@i-like-robots


env.scss
```diff
@@ -15,10 +15,13 @@ $_next-sass-setup_applied: true;
 
 @import 'configure';
 
-@import 'buttons/main';
+// primitives
```
i-like-robots: +1 for terminology =]

PR #218, opened 2016-07-08
Set: o-expander 'is-silent' value to false

andygout:
cc @i-like-robots @keirog 

Looking at this [PR](https://github.com/Financial-Times/n-ui/commit/b8a9ee320c499ced6125bae5a404dbd456cef4a7#diff-02637b46ebfc732031e2457081a8bba7), seems most components have `is-silent` state set to `false` (by prefacing `nUiHas` with `not`), but not `expander` (changed here), and also [`drawer`](https://github.com/Financial-Times/n-ui/blob/1e67440427a27da664f8d9fde10710ec3de4ab5c/drawer/main.scss#L4) (using `header` as its `nUiHas` argument), although that is still working fine.

##### From:
screen shot 2016-07-08 at 11 35 57

##### To (w/ toggle functionality reinstated):
screen shot 2016-07-08 at 11 36 07

### DIFF
--- a/expander/main.scss
+++ b/expander/main.scss
@@ -1,6 +1,6 @@
 @include nUiCriticalStart('expander');
 
-$o-expander-is-silent: nUiHas('expander') !default;
+$o-expander-is-silent: not nUiHas('expander') !default;
 
 @import 'o-expander/main';
 


i-like-robots:
Whoops, good spot! 🚢

andygout:
Do you know why `drawer` is still working while seemingly having silent state set to true?

i-like-robots:
Because it doesn't support silent mode =]

i-like-robots:
When it becomes a part of o-header it will

andygout:
Cool.

PR #221, opened 2016-07-08
Remove `o-header` and `awesomplete` resolutions

leggsimon:
`n-ui` should be the only thing requiring `o-header` in next now so we _shouldn't_ need these resolutions.

cc/ @i-like-robots @wheresrhys


i-like-robots:
+1

wheresrhys:
redundant

PR #225, opened 2016-07-11
replace deprecation-warning n* with deprecation-happy o*

bjfletcher:
Hi there,

The build log for `next-stream-page` has over time got very large, mainly because of deprecation warnings. I thought I'd spend 5 minutes looking at this, and making a start with this PR.

Really the main reason for this is for me to better understand whether:
a) we can replace `n-button` with `o-button` now?
b) how'd I ensure that any changes to `n-ui` won't have an adverse effect on an app somewhere, such as `myFT` - would I need to setup and run all apps that use `n-ui` & test that they're all looking ok? (Although I'm not sure this is what `n-ui` contributors do?)

### DIFF
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -15,19 +15,19 @@
 	}
 
 	.n-myft-ui__button {
-		@include nButtons();
+		@include oButtons();
 	}
 
 	.n-myft-ui__button--inverse {
-		@include nButtonsTheme(inverse);
+		@include oButtonsTheme(inverse);
 	}
 
 	.n-myft-ui__button--medium {
-		@include nButtonsSize(medium);
+		@include oButtonsSize(medium);
 	}
 
 	.n-myft-ui__button--big {
-		@include nButtonsSize(big);
+		@include oButtonsSize(big);
 	}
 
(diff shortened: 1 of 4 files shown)


i-like-robots:
I can answer b) I tend to test test using the streams and article apps rather than all. @wheresrhys did some work on implementing a `make demo` task too which I could try and pick up this week if I have time.

wheresrhys:
`make demo` should work for everyone now (the first iteration accidentally left in a reference to my dev env). The test harness should also make it east to add basic tests to each component too (though, this won't cover sass)

Broadly speaking I think use the same approach to QA for n-ui as you would for other components - after all, it uses semver. The one caveat I have is __don't release a breaking change unless you're committed to rolling it out to all 10 or so of our user facing apps__ ... which I suppose is another way of saying

> Prefer deprecation and backwards compatibility to breaking changes where possible, unless you've got a lot of time to spare

And :+1: for this PR :smile:

wheresrhys:
nudge @bjfletcher

wheresrhys:
nudge nudge @bjfletcher

PR #229, opened 2016-07-11
Remove old welcome message code

i-like-robots:
I don't think the old one will ever be coming back ;-)


i-like-robots:
cc https://github.com/Financial-Times/n-header-footer/pull/308

PR #233, opened 2016-07-11
$$

wheresrhys:


i-like-robots:
+1 because people will think we use prototype

PR #235, opened 2016-07-12
catch 1% of anon users too & light-signup

lc512k:
@adambraimbridge this should catch 1% of anon users (404s on the session api) and anyone seeing a light-signup form

cc @commuterjoy gave up on session endpoint, using 2% of next-style `spoorId`s


lc512k:
(new github font is 😭)

tracking/third-party/mouseflow.js
```diff
@@ -45,7 +46,15 @@ module.exports = function (flags) {
...
-				if (parseInt(lastUuidSegmentHex, 16) % 100 === 0) { // 1%
+				if (parseInt(lastUuidSegmentHex, 16) % 100 === 0) { // 1% of registered & subscribers
+					enableMouseflow();
+				}
+			})
+			.catch(function () {
+
+				// Session API unreachable; most likely anon user
+
+				if (Math.floor(Math.random() * 100) === 0) { // 1% of anon
```
lc512k: Ugh, I don't like that this doesn't stick with the user, though... :-/

adambraimbridge:
I like it. 👍

PR #239, opened 2016-07-12
Use geebee for o-header logos

ironsidevsquincy:
Depends on https://github.com/Financial-Times/o-header/pull/186


i-like-robots:
+1

PR #245, opened 2016-07-13
hack: Re-classify Facebook in app browser on iOS as core

matthew-andrews:
@i-like-robots

### DIFF
--- a/layout/partials/ctm.html
+++ b/layout/partials/ctm.html
@@ -1,7 +1,10 @@
 {{#if @root.flags.javascript}}
 var classes = document.documentElement.className;
 classes = classes.replace(/\bno-js\b/, 'js');
-window.cutsTheMustard = (typeof Function.prototype.bind !== 'undefined');
+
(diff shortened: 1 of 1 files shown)


layout/partials/ctm.html
```diff
@@ -1,7 +1,10 @@
 {{#if @root.flags.javascript}}
 var classes = document.documentElement.className;
 classes = classes.replace(/\bno-js\b/, 'js');
-window.cutsTheMustard = (typeof Function.prototype.bind !== 'undefined');
+
+// HACK: Make FB in-app browser on iOS not cut the mustard, temporarily
+var isFBInAppBrowser = /OS [1-9].* like Mac OS X.*/.test(navigator.userAgent) && /(FBAN|FBAV)/.test(navigator.userAgent);
```
i-like-robots: Maybe simplify to `/OS [1-9].* like Mac OS X/`

matthew-andrews: yup ok

Issue #246, opened 2016-07-13
Configuration


i-like-robots:
It's come up a few times that configuration of n-ui is confusing people. The JavaScript side is quite self-explanatory but the Sass side is not. Originally it was intended that both should follow the same API (as close as possible) but naturally they have diverged.

The `$_n-ui-features` map lists all of the available modules and their default values. There are two confusing aspects to this:

1. 'primitive' modules (those within `env.scss`) are not meant to be configurable
2. The values are not boolean but may be `true`, `false` _or_ `'critical'`

Because of reason 2 it is not possible to sort out 1.

I think something like the following may be clearer: 

```scss
// declare only the configurable features
$_n-ui-features: (
	ads: true,
	buttons: false,
	cookieMessage: false,
	expander: false,
	footer: true,
	forms: false,
	header: true,
	layout: true,
	messagePrompts: false,
	myft: false,
	notification: false,
	overlay: false,
	promoMessages: false,
	welcomeMessage: true
) !default;

// declare default critical features
$n-ui-critical-features: (
	ads, grid, header, layout, normalize, typography, utils
) !default;
```

With usage as following, with the addition of a separate `critical` parameter:

```scss
@include nUiConfigure((
  preset: 'complete',
  features: (
    cookieMessage: true
  ),
  critical: (
    cookieMessage
  )
);
```

i-like-robots:
http://www.sassmeister.com/gist/f7447023fc69b1daa6540732834e9a87

ironsidevsquincy:
Not sure this is making it simpler. Haven't heard any issues around it being currently confusing, but maybe I'm just missing out on the chatter

1) Everything should be configurable, i.e. some pages don't need ads, util, header, etc, so to future-proof, every feature should have the ability to turn off
2) Personally, quite like the current configration values, i.e. a feature having three states, keeps everything in a easy to parse single list

So my 2p is to leave the api as it is

wheresrhys:
I'm not eager for a change either. Current config hasn't had a chance to settle in yet, and we're not sure all the use cases we need to cover have been addressed. Better to react to new requirements as they come in for a while, and then if at the 'end' of it we have something that's a bit of a mess, then I think we can justify a refactor & API tidy

i-like-robots:
@ironsidevsquincy 1. yes, 2. reminds me of eslint settings 
@wheresrhys sounds sensible

This was somewhat reactionary after being on the receiving end of complaints/frustrations a couple of times.

[...]

PR #250, opened 2016-07-13
Refactor configuration mixins

i-like-robots:
This is an internal refactor, nothing like the stuff I was talking about it in #246!

1. standardises internal variable names (I know, meh)
2. removes unused is/not IE8 mixins
3. abstracts configuration and critical mixins out a little and updates doc comments
4. makes declaring presets simpler 

I only made one small change to the API surface which is that the `nUiConfigure` mixin now only takes a preset name and/or a map of features to configure. To enable critical head CSS/variant just update the flags, E.G.:

This:

```scss
$output-critical: true !default;

@import 'n-ui/configure';

@include nUiConfigure(
	$preset: 'discrete',
	$output-critical: $output-critical
);
```

Becomes: 

```scss
$n-ui-output-critical: true;

@import 'n-ui/configure';

@include nUiConfigure(preset: 'discrete');
```

Simply because everything in Sass is global and obfuscating them behind a mixin didn't seem to add extra convenience. Not against adding the options back in, it just didn't fit right in my head at the time.


configure.scss
```diff
@@ -1,9 +1,11 @@
...
+////
 
-$_n-ui-output-critical: false !default;
-
-// default feature values
-$_n-ui-features: (
+/// Configurable features
+/// @type List
+/// @private
+$n-ui-features: (
```
wheresrhys: why remove the `_` - are these intended as public variables now?

wheresrhys:
I think it's neater, so not against the update, though it'd be a breaking change and I think the onus would be on you to roll it out promptly to all our apps; different apps on different APIs is worse than the current one everywhere

configure.scss
i-like-robots: Partly because they're public anyway, partly because some syntax highlighters (I.E. mine) don't like it 😬 - other than fuzzying up the diff it shouldn't make any difference.

i-like-robots:
My plan (if it doesn't offend anybody greatly) is to make the changes to apps _before_ merging this. It's also an option to keep API compatibility but in for a penny in for a pound 💰

i-like-robots:
@wheresrhys I've reinstated the configuration options but added deprecation warnings instead. I will submit PRs to the affected apps over the next few days and remove the option when it's no longer used.

PR #256, opened 2016-07-14
Rhys/boostrap object

wheresrhys:

- refactor to make js bootstrapping more testable
- tests and stronger enforcement of exporting components from the js bundle


grid/index.js
```diff
@@ -17,11 +13,11 @@ function listenForBreakpointChanges(){
 		return;
 	}
 
-	let lastBreakpoint = getCurrentBreakpoint();
+	let lastBreakpoint = oGrid.getCurrentLayout();
 	fireBreakpointEvent(lastBreakpoint);
 	viewport.listenTo('resize');
 	document.body.addEventListener('oViewport.resize', () => {
-		let breakpoint = getCurrentBreakpoint();
+		let breakpoint = oGrid.getCurrentLayout();
```
i-like-robots: ahh this stuff needs cleaning... will submit a PR later

js-setup/js/component-initializer.js
```diff
@@ -0,0 +1,153 @@
...
+		footer: true,
+		date: true,
+		cookieMessage: true,
+		welcomeMessage: true,
+		myft: true,
+		messagePrompts: true,
+		promoMessages: false,
+		ads: true
+	}
+};
```
i-like-robots: Could this should be stored at the top level somewhere, IE. is it too likely to be missed here for people adding/changing components?

js-setup/js/component-initializer.js
```diff
@@ -0,0 +1,153 @@
...
+			}
+
+			if (config.features.date && !this.isInitialized('date')) {
+				date.init();
+				this.initializedFeatures.date = true
+			}
+
+			mainCss
+				.then(() => {
+					if (config.features.cookieMessage && !this.isInitialized('cookieMessage')) {
```
i-like-robots: the method call doesn't seem to add much convenience, maybe it could take a second arg to set? 

```js
this.initialized('cookieMessage') // false
this.initialized('cookieMessage', true) // sets to true
```

wheresrhys: ah yes, thta's a hangover from old hacks to make the code testable. will think about removing it

js-setup/js/component-initializer.js
wheresrhys: Perhaps in another PR reconciling presets, _entry and main.js can be done. I agree that it's lacking a sense of there being a single source of truth

PR #272, opened 2016-07-19
stops us flashing our drawers

wheresrhys:
@aintgoin2goa @i-like-robots 

caveat - not tested it yet


wheresrhys:
Doesn't appear to break anything. As the flashing drawer is intermittent, not sure if it's fixed now

i-like-robots:
This is fixed in the `o-header` implementation which I working on pulling back in right now.

PR #274, opened 2016-07-19
o-header drawer

i-like-robots:
- Integrates the drawer implemented in o-header.
- Refactors opt out JS to shared component
- Puts opt out links behind their associated flag

screen shot 2016-07-19 at 14 51 59

screen shot 2016-07-19 at 14 50 38

Tested in Chrome, iOS 9 and IE9


js-setup/js/component-initializer.js
```diff
@@ -3,6 +3,7 @@ import ads from '../../ads';
 import tracking from '../../tracking';
 import date from '../../date';
 import header from '../../header';
+import optInOut from '../../opt-in-out';
```
aintgoin2goa: why optInOut?  We don't/can't opt people in?

i-like-robots: True - I switched it to match the flag but will switch back

i-like-robots: done

PR #293, opened 2016-07-26
pass meganav section name to beacon

leggsimon:
@i-like-robots 

Passes through the section name to be trackable.

Current: 
screenshot 2016-07-26 11 03 54

Now: 
screenshot 2016-07-26 11 03 18


header/partials/meganav/template.html
```diff
@@ -1,7 +1,7 @@
 {{#if meganav}}
 <div class="o-header__mega" id="o-header-mega-{{@index}}" role="group" aria-labelledby="o-header-link-{{@index}}" data-o-header-mega data-trackable="meganav">
```
i-like-robots: Perhaps it should be `meganav: {{name}}` @adambraimbridge ?

leggsimon: Yeah that seems a bit nicer but I'll wait to see what Braim Storm has to say

i-like-robots:
wünderbar 👍

PR #297, opened 2016-07-26
fix: Add cookies to /search-suggestions request so that the FT_SITE cookie is added so that it works on dubnext

matthew-andrews:
one small dubstep for next


matthew-andrews:
yah good point, i think in this case since `/search-suggestions` isn't used by old FT.com i might make this a Next only route.

PR #295, opened 2016-07-26
Refactor z-index mixin

i-like-robots:
As per the TODO =]

cc @leggsimon @keirog

PR #299, opened 2016-07-26
Fix z index order

leggsimon:
@i-like-robots @keirog 

Puts `typeahead` above `meganav` so it doesn't hide behind the meganav when this happens

screenshot 2016-07-26 14 51 29

And removes `banner`, `sticky-header`, `drawer` use-cases as they are unused.

(related to https://github.com/Financial-Times/n-ui/pull/295)


i-like-robots:
+1

I will re-introduce the drawer and sticky nav use cases when I actually implement them ;-)

PR #300, opened 2016-07-26
Add querystring params to the polyfill service call

leggsimon:
@georgecrawford @bjfletcher @wheresrhys 

This should force unknown user agents to be polyfilled. I haven't added it to the call on line 71 as it is currently always polyfilled.

### DIFF
--- a/js-setup/templates/bootstrapper.html
+++ b/js-setup/templates/bootstrapper.html
@@ -54,7 +54,7 @@
 				'Array.prototype.find|always|gated',
 				'Array.prototype.findIndex|always|gated',
 				'IntersectionObserver'
-			].join(',') + '&excludes=Symbol,Symbol.iterator,Symbol.species,Map,Set');
+			].join(',') + '&excludes=Symbol,Symbol.iterator,Symbol.species,Map,Set' + '&flags=gated&unknown=polyfill');
 		} else {
 			{{#if hasMainCoreJs}}
 			ftNextLoadScript('{{#hashedAsset}}main-core.js{{/hashedAsset}}');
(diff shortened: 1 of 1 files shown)


georgecrawford:
👍

js-setup/templates/bootstrapper.html
```diff
@@ -54,15 +54,15 @@
...
+			].join(',') + '&excludes=Symbol,Symbol.iterator,Symbol.species,Map,Set' + '&flags=gated&unknown=polyfill');
 		} else {
 			{{#if hasMainCoreJs}}
 			ftNextLoadScript('{{#hashedAsset}}main-core.js{{/hashedAsset}}');
 			{{/if}}
 			ftNextLoadScript(polyfillUrl + '?callback=' + polyfillCallbackName + '&features=' + [
 				'default',
 				'HTMLPictureElement|always|gated'
-			].join(','));
+			].join(',') + '&flags=gated&unknown=polyfill');
```
bjfletcher: @JakeChampion @quarterto does this look good to you? What testing should we do before and after? Apart from JS not working when links opened within Facebook app? Thank-you very much :)

JakeChampion: The querystring parameters and values you've added will ensure that useragents unknown to the polyfill-service will be served all polyfills you have asked for and will wrap each polyfill in a feature detection as to not polyfill anything which the useragent has already implemented.

If your issue is with FB in-app browser, we have a PR to add detection for it -- https://github.com/Financial-Times/polyfill-service/pull/775

I'm not sure how this application is setup, it looks to be calling the polyfill-service twice? Is that correct?

/cc @triblondon as he is looking at FB in-app browser.

leggsimon: @JakeChampion The polyfill-service is called differently depending on whether the browser "cuts the mustard" or not. The call on line 62 is for ones that don't. After speaking with Rhys we decided that adding `&flags=gated&unknown=polyfill` _shouldn't_ make a difference here just because the browsers _are_ so old by shouldn't be any harm.

PR #304, opened 2016-07-27
fix: Try to split the chartbeat data by edition

matthew-andrews:


adambraimbridge:
👍 Looking good locally — let's merge it tomorrow morning @matthew-andrews

PR #307, opened 2016-07-27
Rhys/cors proxy support

wheresrhys:
Sister PR for https://github.com/Financial-Times/next-router/pull/33

@adgad this aims to generalise the jsonp-less approach you have for ads API. Can you point me to anything I would need to remove that's in place for ads only

### DIFF
--- a/ads/index.js
+++ b/ads/index.js
@@ -18,7 +18,7 @@ const customTimings = {};
 function getContextualTargetingPromise (appName) {
 	let uuid;
 	let url;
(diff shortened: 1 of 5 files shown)


adgad:
Nice!

I _believe_ it's just https://github.com/Financial-Times/n-ui/blob/master/ads/index.js#L21 and https://github.com/Financial-Times/ads-api/blob/master/server/app.js#L27 can then be removed.

PR #311, opened 2016-07-28
Remove samsung as unsupported

leggsimon:
Now that the polyfill-service is polyfilling unknown browsers this should fix https://jira.ft.com/browse/NFT-518

@wheresrhys @bjfletcher

### DIFF
--- a/js-setup/templates/ctm.html
+++ b/js-setup/templates/ctm.html
@@ -3,34 +3,7 @@
 	var classes = document.documentElement.className;
 	classes = classes.replace(/\bno-js\b/, 'js');
 
-	// HACK: make the following browsers core as they're unsupported by the polyfill service
-	var isSupported = true;
-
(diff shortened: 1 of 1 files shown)


js-setup/templates/ctm.html
```diff
@@ -7,13 +7,19 @@
...
 	{{#if @root.flags.forceCoreExperience}}
-	var unsupported = {
-		'Samsung frankenchromium': function (ua) {
-			var android = /Android/.test(ua);
-			var samsung = /SamsungBrowser/.test(ua);
-			return android && samsung;
-		}
-	};
+
+	// Add any browsers we want to force core experience upon here by adding to
```
wheresrhys: I would just delete these lines entirely. If it breaks github has a revert button :smile:

leggsimon: I was just keeping them in for documentation really...

PR #315, opened 2016-07-29
Drop n-header-footer and integrate subnav from o-header

i-like-robots:
As it says, no changes to the data are required.


header/partials/subnav/template.html
```diff
@@ -0,0 +1,42 @@
...
+								{{name}}
+							</a>
+						</li>
+						{{/currentNav}}
+					</ol>
+
+					{{#if currentNav.children}}
+					<ul class="o-header__subnav-list o-header__subnav-list--children" aria-label="Subsections" data-trackable="subsections">
+						{{#each currentNav.children}}
+						<li class="o-header__subnav-item">
```
keirog: Do we need the unselected modifier? Surely any default values should just be on the base `o-header__subnav-item` class, no?

keirog: Oh, it's gone

keirog:
👍

PR #322, opened 2016-08-02
align prefgroup with floats not inline-block

laurieboyes:
@keirog 

This is to get around the problem caused by the following facts:

- For the display inline block and margin -5px approach, there needs to be whitespace in between the components
- React strips whitespace from around components, and the solution is to insert some whitespace and wrap the whole thing in a span [for example](https://github.com/Financial-Times/next-myft-page/blob/2c88d7ce76ddddebfc1ca75a305666f5dcf80f6b/shared/components/alerts/preferences/prefer-group.js#L15-L29)
- The pref group styling requires the buttons to not be nested in spans

### DIFF
--- a/myft/scss/_preference.scss
+++ b/myft/scss/_preference.scss
@@ -117,14 +117,18 @@
 		}
 
 		.n-myft-ui__button {
+			@include clearfix();
 			border-radius: 0;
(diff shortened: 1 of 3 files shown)


keirog:
👍

PR #329, opened 2016-08-08
Set the nlayout ad parameter if in the lazy load MVT test

adgad:
/cc @GlynnPhillips @tom-parker @andrewgeorgiou1981 @VladDubrovskis 

Adds the ab state to the HTML tag, and then uses that to pass the state of the lazy load MVT to the ad call.


VladDubrovskis:
👍

PR #336, opened 2016-08-09
Add: Scroll depth component tracking

andygout:
cc @andygnewman @wheresrhys 

Rhys - based on the [documentation](https://github.com/Financial-Times/n-ui/blob/master/README.md), would this constitute a patch (`adding a js method`) or minor (`js to be updated at the same time as the templates`, although the app's JS does not necessarily need to be updated at the same time) release (so I am guessing the former)?


wheresrhys:
yep go for a patch

andygout:
Is that a 👍 ?

tracking/ft/events/utils/fire-tracking.js
```diff
@@ -0,0 +1,10 @@
+export default (name, data, el = document.body) => {
```
wheresrhys: You can use https://github.com/Financial-Times/n-ui/blob/master/utils/index.js#L45 instead of defining this helper

PR #340, opened 2016-08-11
Add full page location as a referrer param on the opt-out link

TheoLeanse:
/cc @matthew-andrews @adambraimbridge 

By using this parameter rather than the value of the `Referer` header (used [here](https://github.com/Financial-Times/next-opt-in-out/blob/master/server/controllers/opt-out-form.js#L20)) we may forever banish the unpredictable redirections when landing back on Falcon after opting out of Next FT.com (and thereby satisfy [this jira ticket](https://jira.ft.com/browse/NFT-534))

### DIFF
--- a/opt-out/index.js
+++ b/opt-out/index.js
@@ -18,6 +18,10 @@ function showAndroidLink () {
 }
 
 function init () {
+	$$('.js-optout-link').forEach(a => {
+		const referrer = encodeURIComponent(location.pathname);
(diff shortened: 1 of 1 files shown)


adambraimbridge:
👍

matthew-andrews:
hm, can we not do this on the server side?

matthew-andrews:
otherwise our core experience will continue to be unpredictable

PR #343, opened 2016-08-11
Use nice link from n-navigation

TheoLeanse:
/cc @aintgoin2goa @i-like-robots @matthew-andrews @adambraimbridge 

and it's definitely `navigation.lists` not `navigationLists`?


i-like-robots:
+1

aintgoin2goa:
They both work - but navigationLists is depreacted

adambraimbridge:
👍

leggsimon:
+1

PR #344, opened 2016-08-15
Ability to disable rendering of welcome message

ironsidevsquincy:
With `withWelcomeMessege` param

### DIFF
--- a/layout/wrapper.html
+++ b/layout/wrapper.html
@@ -87,7 +87,7 @@
 			<div class="n-layout__row n-layout__row--footer">
 				{{#outputBlock 'footer'}}{{/outputBlock}}
 				{{#if @root.flags.welcomePanel}}
-					{{#ifEquals withWelcomeMessage false}}
+					{{#ifEquals nUi.welcomeMessage false}}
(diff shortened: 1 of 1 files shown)


wheresrhys:
can this be done as a map of properties following the same naming conventions as n-ui i.e `{{if @root.nUi.welcomeMessage}}`

ironsidevsquincy:
Good idea

layout/wrapper.html
```diff
@@ -87,7 +87,10 @@
 			<div class="n-layout__row n-layout__row--footer">
 				{{#outputBlock 'footer'}}{{/outputBlock}}
 				{{#if @root.flags.welcomePanel}}
-					{{>n-ui/welcome-message/template}}
+					{{#ifEquals nUi.welcomeMessage false}}
```
i-like-robots: Could use `{{#unless ... }}` maybe?

ironsidevsquincy: Unfortunately not, as that would mean updating every app that uses this template (as `undefined` would count as a positive)

PR #358, opened 2016-08-19
feat(tracking): track segmentID if cookie is present

keithamus:
This change detects if the `segmentID` cookie is present, and, if so, adds it to the o-tracking context object, so that it is tracked for all events on the page. segmentID is set as a session cookie, and is not always present.

segmentID is used for campaign tracking for inbound ads, through - for example - facebook ads. We need to track segmentID so we can attach anonymous user journeys to particular campaigns.

Based on git blame data, I'm going to ping @wheresrhys, @GlynnPhillips and @commuterjoy for anyone who fancies reviewing. I'm also going to ping @ironsidevsquincy @matthew-andrews who I've spoken to about the requirements for this. @matthew-andrews also suggests @i-like-robots as a potential reviewer.

### DIFF
--- a/tracking/ft/index.js
+++ b/tracking/ft/index.js
@@ -64,6 +64,10 @@ const oTrackingWrapper = {
 
 			const edition = document.querySelector('[data-next-edition]') ? document.querySelector('[data-next-edition]').getAttribute('data-next-edition') : null;
 			context.edition = edition;
+			const segmentID = String(document.cookie).match(/(?:^|;)\s*segmentID=([^;]+)/) || [];
+			if (segmentID[1]) {
+				context.segmentID = segmentID[1];
+			}
 
 			oTracking.init({
 				server: 'https://spoor-api.ft.com/ingest',


wheresrhys:
Looks OK to me. You may want to check that the segID also gets added to tracking of any click events etc (not sure if that falls through automatically from  oTracking init)

wheresrhys:
:+1:

tracking/ft/index.js
```diff
@@ -64,6 +64,9 @@ const oTrackingWrapper = {
 
 			const edition = document.querySelector('[data-next-edition]') ? document.querySelector('[data-next-edition]').getAttribute('data-next-edition') : null;
 			context.edition = edition;
+			if (/(?:^|;)\s*segmentID=([^;]+)/.test(document.cookie)) {
+				context.segmentID = RegExp['$1'];
```
ironsidevsquincy: Ohhh, didn't know about this. Is it safe? (just going by the big 'Non-standard warning at the top of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/n)

keithamus: I'll refactor.

PR #360, opened 2016-08-22
Load chartbeat for a random 20% of device IDs

adambraimbridge:
Thanks to @lc512k for the 20% code example

/cc @commuterjoy


tracking/third-party/chartbeat.js
```diff
@@ -24,12 +24,18 @@ const enableChartbeat = () => {
...
-		if (flags.get('chartbeatHud') || Math.random() < 0.2) {
+	if (flags.get('chartbeatHud')) {
+		enableChartbeat();
+	}
+	else if (flags.get('chartbeat')) {
+		const spoorId = getCookieValue('spoor-id');
+		const spoorNumber = spoorId.replace(/-/g, '');
+		const spoorNumberTrim = spoorNumber.substring(spoorNumber.length - 12, spoorNumber.length); // Don't overflow the int
+		const spoorNumberDec = parseInt(spoorNumberTrim, 16);
+		if (spoorNumberDec % 10 < 20) { // 20%
```
lc512k: @adambraimbridge, `% 100` rather than `%10` I think?

Also with mouseflow I've found that even `% 100 === 0` turned out to be significantly more than 1% (throttled it in the mouseflow site further), which makes me think the spoorIds aren't completely random, or we're skewing it by truncating and/or parsing to decimal...

adambraimbridge:
Currently blocked by https://github.com/Financial-Times/n-webpack/pull/24

PR #364, opened 2016-08-22
feat(segments): hide wecome message if coming from segment

keithamus:
This change will intentionally hide the "welcome to next" message if the user has a segmentId cookie.

Users with segmentId have come from inbound ad channels, and are forced into next - and as such the opt-in notification does not fit with their user journey. There is no journey on Falcon for these users, so they have to be on next - we should avoid telling them they can have the same experience by opting out (because they wont).

Blame log says @wheresrhys @i-like-robots should check this out. Also paging @matthew-andrews @ironsidevsquincy who've been briefed on the business requirements for this change.

### DIFF
--- a/welcome-message/index.js
+++ b/welcome-message/index.js
@@ -17,6 +17,15 @@ function init () {
 	const fixedEl = $('.n-welcome-message--fixed');
 	const staticEl = $('.n-welcome-message--static');
 
+	const segmentId = String(document.cookie).match(/(?:^|;)\s*segmentID=/);
+	if (segmentId) {
+		if (hasLocalStorage()) {
+			superstore.local.set(STORAGE_KEY, 1);
+		}
+		fixedEl.hidden = true;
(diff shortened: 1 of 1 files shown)


ironsidevsquincy:
Might be best to not event output the template into the page at all (https://github.com/Financial-Times/n-ui/blob/master/layout/wrapper.html#L90)
Maybe a middleware that doesn't add it if the cookie's present (will mean varying on presence of segmentId)

keithamus:
Hmm... as we discussed earlier, adding vary for segmentid probably breaks up cache too much right?

keithamus:
I guess, as you say, we could break it up based on _presence of segmentid_

keithamus:
Okay, face to face discussion we have settled on this staying as JS, as its temporary and we don't want to break up cache more.

ironsidevsquincy:
👍

PR #365, opened 2016-08-22
Throw a hard error (bail) if webpack fails

adambraimbridge:
Throw a hard error (bail) if webpack fails to ensure that make test doesn't have a false positive.

/cc @ironsidevsquincy


ironsidevsquincy:
👍

PR #371, opened 2016-08-24
Tracking - pick page-meta from document & add to page view context

andygnewman:
cc: @wheresrhys @adambraimbridge

Give the capability for apps to render a JSON object into the document tagged with `data-page-meta` that can then be added to the context property of the page view dispatched via oTracking.

### DIFF
--- a/tracking/ft/index.js
+++ b/tracking/ft/index.js
@@ -68,6 +68,15 @@ const oTrackingWrapper = {
 			if (segmentId[1]) {
 				context['marketing_segment_id'] = segmentId[1];
 			}
+			const pageMetaEl = document.querySelector('[data-page-meta]');
+			let pageMeta;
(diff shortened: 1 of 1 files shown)


tracking/ft/index.js
```diff
@@ -68,6 +68,15 @@ const oTrackingWrapper = {
 			if (segmentId[1]) {
 				context['marketing_segment_id'] = segmentId[1];
 			}
+			const pageMetaEl = document.querySelector('[data-page-meta]');
+			let pageMeta;
+			if (pageMetaEl) {
+				try {
+					pageMeta = JSON.parse(pageMetaEl.innerHTML);
```
wheresrhys: why not make the pageMetaEl a script tag?

wheresrhys:
:+1:

PR #380, opened 2016-08-30
Instant alerts core

debugwand:
* update-myft-form is aimed to give us a template where we pass in data and it renders any button to update myft with
* updates to n-ui to come later
* no tests yet because karma-webpack-enzyme is throwing unexpected errors


keirog:
👍

PR #391, opened 2016-09-02
Alter name and value handling

debugwand:


pixelandpage:
👍

debugwand:
Tests completed succesfully, however someone else keeps pushing to master without allowing their tests to complete, I have rebased and re-ran tests twice already, hence last item being yellow and not a green tick.

PR #401, opened 2016-09-06
improved dev workflow and docs

wheresrhys:


Makefile
```diff
@@ -20,17 +20,22 @@ test-server:
...
 	node ./_deploy/s3.js
 	$(MAKE) npm-publish
 
-test: verify test-unit test-build test-server nightwatch
+build:
+	export DEV_BUILD=1; webpack
 
-test-dev: verify test-unit-dev
+watch:
+	export DEV_BUILD=1; webpack --watch
```
adambraimbridge: Could this be `watch: build --watch` ..?

wheresrhys: unfortunately not. `make` only understands task names and not bash options so that would throw an error as `--watch` isn't a valid name for a task

PR #404, opened 2016-09-06
Enhanced instant alert

debugwand:


myft/ui-instant.js
```diff
@@ -0,0 +1,121 @@
...
+const UI_HOOK = '[data-myft-ui="instant"]';
+let config;
+
+// until API is updated to return modelled response data from create calls, fallback to old raw format
+function apiBackwardsCompatibility (response) {
+	return (response._rel) ? response._rel.instant : response.results[0].rel.properties.instant;
+}
+
+function updateConceptData (actorId, subjectId, data) {
+	myftClient.updateRelationship('user', actorId, 'followed', 'concept', subjectId, data);
```
phamann: To avoid passing `actorId` and fetching from the DOM you can pass `null` here and the client will use the memoized uuid it has from the initial init.

phamann:
👍 LGTM

PR #408, opened 2016-09-07
allow app to turn off sticky header

tavvy:
/cc @i-like-robots

### DIFF
--- a/header/template.html
+++ b/header/template.html
@@ -34,7 +34,9 @@
 	{{/unlessEquals}}
 </header>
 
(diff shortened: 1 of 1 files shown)


header/template.html
```diff
@@ -34,7 +34,9 @@
 	{{/unlessEquals}}
 </header>
 
-<header class="o-header o-header--simple o-header--sticky o--if-js" aria-hidden="true" role="presentation" data-o-header--sticky>
-	{{> n-ui/header/partials/header/sticky}}
-	{{> n-ui/header/partials/header/search context="sticky"}}
-</header>
+{{#unless nUi.header.noSticky}}
```
i-like-robots: maybe `disableSticky`? negative thing == true seems weird

PR #409, opened 2016-09-07
consume n-image to enable putting lazy loading in here

wheresrhys:
@ironsidevsquincy


wheresrhys:
see also https://github.com/Financial-Times/next-front-page/pull/1024

PR #411, opened 2016-09-07
Myft digest promo component

pixelandpage:
Displays a promo banner for the email digest on article stream pages. CTA adds topic to the user's myFT and subscribes them to the email digest.

screen shot 2016-09-07 at 17 21 22


my-ft-digest-promo/index.js
```diff
@@ -0,0 +1,86 @@
...
+  }, false);
+}
+
+function shouldShowPromo(conceptId) {
+  return Promise.all([
+    myFtClient.get('followed', 'concept', conceptId),
+    myFtClient.get('preferred', 'preference', 'email-digest')
+  ]).then(values => {
+    return values[0].length === 0 && values[1].length === 0 && !getDismissState();
+  });
```
laurieboyes: This bit could be tidied up a bit by making use of [destructuring](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). e.g. 
```
.then([follows, prefers] => {
   return follows.length === 0 && prefers.length === 0 && !getDismissState();
})
```
(Also destructuring gets you mad hipster points)

my-ft-digest-promo/promo-message.html
```diff
@@ -0,0 +1,38 @@
...
+				<button class="n-myft-digest-promo__dismiss-btn" data-trackable="digest-promo-dismiss"></button>
+			</div>
+		</div>
+	</div>
+
+		<div class="o-grid-container n-myft-digest-promo__container">
+			<div class="o-grid-row">
+				<div data-o-grid-colspan="12 M8 L10" class="n-myft-digest-promo__content">
+					<h2 class="n-myft-digest-promo__heading">
+					  Add to <nobr><span style="font-weight:normal">myFT Digest</span></nobr>
```
laurieboyes: Oh wow TIL what a `<nobr>` is. MDN suggests it's best to go for `white-space: nowrap` over this though: https://developer.mozilla.org/en/docs/Web/HTML/Element/nobr

my-ft-digest-promo/promo-message.html
```diff
@@ -0,0 +1,38 @@
...
+					</h2>
+					<p class ="n-myft-digest-promo__para">Add this topic to a myFT, straight to your inbox</p>
```
laurieboyes: Love the BEM mastery 👍

myft-digest-promo/index.js
```diff
@@ -0,0 +1,93 @@
+const myFtClient = require('next-myft-client');
```
laurieboyes: This file seems to be a duplicate of `my-ft-digest-promo/index.js`? (Although `myft-digest-promo` is the better folder name 👍)

myft-digest-promo/promo-message.html
```diff
@@ -0,0 +1,46 @@
+<!-- Signed in
```
laurieboyes: Dupe again

myft-digest-promo/main.scss
```diff
@@ -0,0 +1,100 @@
+$icon-width-s: 90px;
+$icon-width-m: 90px;
+
+.n-myft-digest-promo {
+  @include clearfix();
+  @include oGridColspan($span: 12);
+
+  display: none;
+  position: relative;
+  float: none;
```
laurieboyes: Don't think this line is necessary, should be default (?)

myft-digest-promo/main.scss
```diff
@@ -0,0 +1,100 @@
...
+  background-color: #fdf8f2;
+  border-bottom: 1px solid #cec6b9;
+  border-top: 1px solid #cec6b9;
```
laurieboyes: Rather than setting hash codes, we want to be pulling colours with `getColor` or `getColorFor` if any of them fall into the use cases (I haven't looked into whether they do or not, sorry 😶): https://github.com/Financial-Times/n-ui/tree/master/colors

Here's all the origami colours we have defined: http://registry.origami.ft.com/components/o-colors@3.5.2

If the colours we need aren't in there then, uh, I guess, try and check with a designer that we can switch it to a similar origami colour? Or if there's justification enough maybe we can add a new one to n-ui (this might need more discussion).

laurieboyes:
A few comments but in general, looks great! 👍 ✨

ironsidevsquincy:
Will this component just be displayed for signed in users?

phamann:
> Will this component just be displayed for signed in users?

Correct. And your LSU I assume is only shown for annon users. So there is no conflict :wink:

ironsidevsquincy:
Yep, that's what I was getting at. Thanks

PR #412, opened 2016-09-08
Fix myft promo default float to show buttons

pixelandpage:
The `oGridColspan` introduced an `float:left` which broke the button flow. Having to override to fix the issue.


phamann:
👍

PR #416, opened 2016-09-09
Update mouseflow.js

lc512k:
Looking at this with pause I noticed the light signup is what's driving up the numbers here @adambraimbridge 

So now we capture:
- signup views (~150 a day)
- random 1% of light signup views (~1200 a day)
- 1% of _visitors_ by spoorid (~1000 a day)

If those calculations are correct, we'll be fairly below our mouseflow quota, so I'll test this out for a bit and increase #3 up to near capacity. This will hopefully help us get better recordings results in Beacon!

### DIFF
--- a/tracking/third-party/mouseflow.js
+++ b/tracking/third-party/mouseflow.js
@@ -21,7 +21,7 @@ module.exports = function (flags) {
 		if (isSignUpApp) {
 			enableMouseflow();
 		}
-		if (hasLightSignup && Math.round(Math.random() * 100) === 1) {
+		if (hasLightSignup && Math.random() <= 0.01) {
 			enableMouseflow();
 		}
 		else {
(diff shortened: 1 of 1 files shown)


lc512k:
This is generating about 50 recordings a minute, which is better than before but still a lot higher than I expected. We're still curbing it in mouseflow a bit, though I want to remove the filter there soon and control everything in n-ui. I'll stay on this for the next few hours @commuterjoy @adambraimbridge

lc512k:
Down to ~5 a minute now, which is more like what we want. No more throttling on the mouseflow UI 🎉 

Not sure why it took a while to stabilize. The new version of https://next-geebee.ft.com/n-ui/no-cache/v2/es5-core-js.min.js was live over two hours ago, though I'm not certain how we handle caching for it.

adambraimbridge:
🎉🎉🎉🎉

PR #419, opened 2016-09-12
Tour tips

laurieboyes:
@keirog's lovely tour tip work (in progress)

Also a getRandomBySize() function for use by the homepage and presumably other pages

e.g.
screen shot 2016-09-12 at 17 06 57


tour-tip/config.js
```diff
@@ -0,0 +1,51 @@
+export default [
```
laurieboyes: Changed this from a `.json` to a `.js` because I was having a nightmare using it in tests (even though really I should be mocking it out. More on this below.)

tour-tip/test/get-tip-model.spec.js
```diff
@@ -0,0 +1,34 @@
...
+			expect(getById('???')).to.be.undefined;
+		});
+
+		it('is the default function', () => {
+			expect(getTipModelDefault === getById).to.be.true;
+		});
+
+	});
+
+	describe('getRandomOfSize', () => {
```
laurieboyes: _Very_ weak testing on this function but marginally better than no testing at all. Wasn't able to successfully mock out the config file, despite banging my head against my keyboard for some time :confounded:

tour-tip/component.js
```diff
@@ -0,0 +1,40 @@
+import React, { Component } from 'react';
+import classNames from 'classnames';
+import { Image } from '@financial-times/n-image';
+import { buildImageServiceUrl } from '@financial-times/n-image/src/helpers';
+
+export default class TourTip extends Component {
+	render () {
+		let attrs = Object.assign({}, this.props.data.data);
```
debugwand: should the props have been passed in as {...data}?

tour-tip/component.js
```diff
@@ -0,0 +1,40 @@
...
+
+		const tourLink = (attrs.settings.hasTourPageLink) ?
+			(<p className='tour-tip__link'><a href='/tour' className='tour-tip__link' data-trackable='tour-link'>New hints and tips</a></p>)
+			: null;
+
+		const body = (attrs.data.body) ?
+			(<p className='tour-tip__body'>{attrs.data.body}</p>) : null;
+
+		return <aside className={classNames(classes)} data-trackable={`tour-tip-${attrs.id}`}>
+			<div className='tour-tip__main'>
```
debugwand: shouldn't this be double quotes as it is to be rendered as HTML?

tour-tip/get-tip-model.js
```diff
@@ -0,0 +1,24 @@
+import tipsConfig from './config';
+
+function deepCopy(obj) {
+	return JSON.parse(JSON.stringify(obj));
+}
+
+export function getById(id) {
+	const tipFound = tipsConfig.find(tip => tip.id === id);
+	if (!tipFound) {
```
debugwand: you could also have if tipFound return deepCopy(tipFound) rather than checking for a negative result first, because a function that doesn't return anything will return undefined

tour-tip/get-tip-model.js
```diff
@@ -0,0 +1,24 @@
...
+		return;
+	}
+	return deepCopy(tipFound);
+}
+
+export function getRandomOfSize (size) {
+	const tipsOfSize = tipsConfig.filter(tip => tip.settings.size === size);
+	if(!tipsOfSize.length) {
```
debugwand: as above

tour-tip/test/get-tip-model.spec.js
debugwand: how about this approach? https://sazzer.github.io/blog/2015/08/20/Unit-Testing-ES6-Modules-Mockery/

tour-tip/component.js
keirog: Oops, my bad, thanks. Will amend

keirog:
@Financial-Times/next-myft Hey team, any further comments before I merge? (It’s all still a WIP and behind a flag.)

laurieboyes:
👍  woop

PR #422, opened 2016-09-13
Ads1024 - Sticky Right-rail ad

andrewgeorgiou1981:


ads/index.js
```diff
@@ -118,12 +118,22 @@ module.exports = {
...
-
+				if(flags && flags.get('stickyHeaderAd')) {
+					let stickyAd = new Sticky(
+						document.querySelector('.above-header-advert'),
+						{ 'sibling' : '.header-ad-placeholder__top',
+						'stickUntil' : '#primary-nav .o-header__top'
+						});
+					stickyAd.init();
+				}
+//				if(flags && flags.get('stickyRightAd')) {
```
adgad: Should this be commented out?

andrewgeorgiou1981: not sure why this is commented out - it seems that some commits might not have been pushed up correctly - (i think there is a githook which runs verify and prevents pushing if things are'nt passing lint).
I'll push the latest up

PR #427, opened 2016-09-14
unregistering service worker inline

wheresrhys:
@richard-still-ft


wheresrhys:
See also https://github.com/Financial-Times/n-service-worker/blob/master/unregister.html

PR #434, opened 2016-09-15
Page blacklist for the myFT digest promo

laurieboyes:
Don't know how happy people will be with me gaffa-taping n-ui, but currently the [us-election-2016 stream page](https://www.ft.com/us-election-2016) has an extra component on it, and that component is overriding some origami styles that are depended on by the myFT promo.

Hopefully this will be a temporary measure while we get the root cause resolved.

It should look like this:
screen shot 2016-09-15 at 16 53 11

But on the us elections page it currently looks like this:
screen shot 2016-09-15 at 16 53 21

### DIFF
--- a/myft-digest-promo/index.js
+++ b/myft-digest-promo/index.js
@@ -10,6 +10,10 @@ const CLASSES = {
 	promoEnabled:'n-myft-digest-promo--enabled'
 };
 
+const pageBlacklist = [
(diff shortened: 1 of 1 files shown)


laurieboyes:
Thing this is a bad idea after all, going to try and fix it properly tomorrow

PR #440, opened 2016-09-16
added maybe alias to nUiCritical to remind users that it's configurable

wheresrhys:

In the next version I reckon we should have the version without `maybe` not accepting variables


ironsidevsquincy:
Was this from feedback? IMO was obvious previously and seems a bit more confusing (and verbose) now

PR #441, opened 2016-09-16
Make our styles beat the us-election info styles

laurieboyes:
There's a component on the [us-election-2016 stream page](https://www.ft.com/us-election-2016) that has its own style tag. The style tag brings in o-grid and tramples some of the myFT promo styles. It also sets a font size and weight `h2` tags, which the myFT promo was relying on.

This PR nests the myFT promo styles in the parent element class, helping it win the specificity fight (two classes beats defined-lower-down).

It also explicitly sets the myFT promo heading font so it's not affected by the h2 overrides.

Before:
screen shot 2016-09-16 at 14 41 59

After:
screen shot 2016-09-16 at 14 42 06


debugwand:
looks good. slack isn't picking up on new shiny changes approved status yet :(

PR #442, opened 2016-09-16
ensure all code before bootstrap has no dependency on polyfill

wheresrhys:
@jakechampion - I think this nails the polyfill issue I had in ie


JakeChampion:
Nice :-)

wheresrhys:
ah, still broken in ie9 though. I think _that_probably is a polyfill issue, though hopefully I'm wrong

PR #444, opened 2016-09-19
Enhanced welcome banner

debugwand:


welcome-message/banner.html
```diff
@@ -1,19 +1,21 @@
-<div data-o-component="o-expander" class="o-expander o-expander__content--expanded n-welcome-banner__wrapper" data-o-expander-shrink-to="hidden">
-	<h6 class="n-welcome-banner__heading n-welcome-banner__column n-welcome-banner__column--m">Welcome to the new FT.com.</h6>
-
-	<div class="o-expander__content n-welcome-banner__column n-welcome-banner__column--fill">
+<div class="n-welcome-banner__wrapper">
+	<h6 class="n-welcome-banner__heading n-welcome-banner__column">Welcome to the new FT.com.</h6>
```
laurieboyes: Guess this is more of a copy question (sorry) - do we definitely want the full stop at the end of this? The 'Ft dot com dot' jars a bit with me

welcome-message/index.js
```diff
@@ -1,22 +1,103 @@
...
+let fixedElHeight;
+let staticEl;
+let viewportHeight;
+
 function hasLocalStorage () {
 	superstore.local.set(TEST_KEY, TEST_VAL);
 	const retrieved = superstore.local.get(TEST_KEY);
 	superstore.local.unset(TEST_KEY);
 	return TEST_VAL === retrieved && superstore.isPersisting();
 }
```
phamann: I know not part of your refactor, but `superstore` actually has an `isPersisting()` method which could replace this function. https://github.com/matthew-andrews/superstore-sync#ispersisting

welcome-message/index.js
```diff
@@ -1,22 +1,103 @@
...
+	});
+}
+
+// in addition to what o-expander gives us, add a class onto the container when changing expanded state
+function toggleContainerDisplay (showFull) {
+	if (showFull) {
+		fixedEl.setAttribute('class', fixedEl.className.replace(/\b\sn-welcome-banner--collapsed\b/g, ''));
+	} else {
+		fixedEl.setAttribute('class', fixedEl.className += ' n-welcome-banner--collapsed');
+	}
```
phamann: I'd use `classList.toggle()` here instead of regex and string manipulation.

welcome-message/main.scss
```diff
@@ -1,15 +1,16 @@
...
 	@include nUiMaybeCriticalStart('welcomeMessage');
 
-	.n-banner-message,
+	.n-welcome-banner,
 	.n-welcome-message {
 		position: relative;
 		padding: 16px 12px;
 		// not in o-colors ಠ_ಠ
-		background: #FADBCB;
+		background: #fadbcb;
```
phamann: Why not an o-colours colour?

welcome-message/main.scss
```diff
@@ -57,29 +79,24 @@
...
 	}
 
-	.n-banner-message {
-		background: getColor('teal-1');
-		color: getColor('white');
-	}
-
 	.n-welcome-banner__heading {
+		display: none;
+		@at-root .n-welcome-banner--collapsed & {
```
phamann: Nice!

welcome-message/index.js
debugwand: is that polyfilled? because natively its IE10+ only

welcome-message/index.js
```diff
@@ -25,19 +106,35 @@ function init () {
...
+}
 
+// this is the 'old' welcome functionality, where you only saw the sticky welcome once
+// we might go back to this once users are confident with new site
+function initOneTimeSticky () {
+	hideIfSegmentId();
 	if (Boolean(superstore.local.get(STORAGE_KEY)) === false && hasLocalStorage()) {
-		const closeButton = $('button', fixedEl);
-
+		const closeButton = fixedEl.querySelector('button');
```
laurieboyes: I'm not sure if we're restricted by o-expander in this but could we use a class selector here instead of a tag one? Maybe make a new one called `js-close-button` or something so it's decoupled from layout and markup implementation?

welcome-message/main.scss
laurieboyes: any chance we could swap this out to something acceptably close that is in o-colors?

debugwand: that's the old welcome message pink bg - @i-like-robots might know

welcome-message/main.scss
```diff
@@ -34,10 +42,24 @@
...
 		position: fixed;
 		right: 0;
-		bottom: 0;
+		bottom: -1px;
 		left: 0;
 	}
 
+	.n-welcome-banner--collapsed {
+		left: auto;
+		right: $spacing-unit/2;
```
laurieboyes: ❤️  spacing-unit

welcome-message/index.js
laurieboyes: the polyfills service doesn't polyfill it for IE10 and 11 for some reason 😢  http://polyfill.webservices.ft.com/v2/docs/features/#Element_prototype_classList

welcome-message/index.js
debugwand: I think it defaults to true and gets updated when you get or set

https://github.com/matthew-andrews/superstore-sync/blob/master/lib/superstore-sync.js#L14

https://github.com/matthew-andrews/superstore-sync/blob/master/lib/superstore-sync.js#L128

welcome-message/main.scss
debugwand: This is the old banner background, nearest valid colour is washed out on our background.
screen shot 2016-09-19 at 12 16 43

welcome-message/index.js
debugwand: This is the old banner code that I have moved into its own function so I can delete it easily.

debugwand: ah - the change was to remove util as $ doesn't give me anything that native JS can't do

welcome-message/banner.html
debugwand: Layout issues:

In Chrome
screen shot 2016-09-19 at 12 19 22

In IE
ie

laurieboyes: Ah yeah I see, it's definitely worth having them  👍

welcome-message/index.js
laurieboyes: Oh yes I see now. Yeah fair enough 👍

welcome-message/main.scss
laurieboyes: Dang 👍

welcome-message/index.js
i-like-robots: We (inc. Origami) use `classList` extensively so it _should_ be polyfilled!

welcome-message/main.scss
i-like-robots: Sometimes we have to do bad things... I don't remember why this is so I'm afraid.

Moving into o-colors would make sense if we wish to use it more widely (though none of the `pink-` colours are actually, currently pink =S)

welcome-message/index.js
```diff
@@ -1,22 +1,103 @@
 const superstore = require('superstore-sync');
-
-import { $ } from '../utils';
+const expander = require('o-expander');
```
i-like-robots: I think o-toggle may suit your needs better in this case

i-like-robots: http://registry.origami.ft.com/components/o-toggle

welcome-message/index.js
debugwand: Had a look at the polyfill, will use `classList` and its `add` and `remove` functions because no version of IE supports the second argument of toggle.

welcome-message/index.js
debugwand: I started with this and switched to o-expander, iirc for the event handlers and because toggler doesn't necessarily hide things

PR #454, opened 2016-09-20
If user has taken tour, don't show stikcy welcome again

debugwand:


welcome-message/index.js
```diff
@@ -108,6 +113,13 @@ function hideIfSegmentId () {
 	}
 }
 
+function setTourButton () {
+	const tourButton = fixedEl.querySelector('[data-component="cta-take-tour"]');
+	tourButton.addEventListener('click', () => {
```
keirog: Won't this element be missing if the flag is off?

debugwand: it won't get called unless `fixedEl.getAttribute('data-component') === 'welcome-banner'` is true

PR #457, opened 2016-09-21
tweak to the wording on the myFT promo banner

pixelandpage:
Small copy change on the myFT promo banner.

Before: 
screen shot 2016-09-21 at 15 14 05

After:
screen shot 2016-09-21 at 15 15 04


debugwand:
looks good

PR #463, opened 2016-09-21
Disable page-view event inside iframes

paddyo:
## Description
Disables the automatically fired `page:view` event if the page is not the top most window. This is to prevent the event from firing in iframes, where doing so will effectively double the recorded traffic and other misleading information

### DIFF
--- a/tracking/ft/index.js
+++ b/tracking/ft/index.js
@@ -94,7 +94,10 @@ const oTrackingWrapper = {
 			};
 
 			// FIXME - should not fire on barriers, but needs to be around for a while data analytics fix their SQL
-			oTracking.page(pageViewConf.context);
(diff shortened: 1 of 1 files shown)


wheresrhys:
this needs to not apply to error pages (which are always served in an iframe)

paddyo:
Does this mean error pages report twice the amount of traffic?

wheresrhys:
Nope, as the error page is in an iframe in a blank page https://github.com/Financial-Times/ft.com-cdn/blob/master/src/vcl/next-errors.vcl (horrible hack, but the best solution we've found for serving error pages)

paddyo:
Is this update, using the error domain, enough?

wheresrhys:
Very likely, but keep an eye on it :+1:

PR #465, opened 2016-09-22
Moar tour amends

keirog:
- Use latest SVGs
- Add intro and outro tip data (for the full page tour, but keeping everything together seems a good idea)
- Stop nudging images to the left (as agreed with design)
- Include tour files in bootstrap (we've been including them in the apps, which was wrong)

### DIFF
--- a/tour-tip/config.json
+++ b/tour-tip/config.json
@@ -9,7 +9,7 @@
     "content": {
       "title": "Track the topics most important to you all in one place",
       "imageAlt": "Illustration of the \"Companies\" page, showing an \"Added\" button next to the page heading",
-      "imageUrl": "https://next-geebee.ft.com/assets/tour/myft-1.svg",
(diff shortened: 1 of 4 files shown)


tour-tip/config.json
```diff
@@ -148,7 +148,7 @@
     "content": {
       "title": "Get intuitive markets data at a glance on any device",
       "imageAlt": "Illustration of a graph in a Markets Data page",
-      "imageUrl": "https://next-geebee.ft.com/assets/tour/markets-1.svg",
+      "imageUrl": "https://next-geebee.ft.com/assets/tour/markets-3.svg",
```
pixelandpage: Did we mean to go from `markets-1` to `markets-3`?

keirog: Ah nope, human error! The file exists though so all good

PR #489, opened 2016-09-26
Copy change on welcome banner

debugwand:


debugwand:
I also do, let's see if it gets clicked

PR #497, opened 2016-09-27
Avoid repaints with fixed elements

debugwand:


phamann:
download

wheresrhys:
Nice diff 
image

PR #501, opened 2016-09-27
Tour image changes

keirog:
- Make the tour intro image responsive (forgot before)
- Change the image to the supplied pack shot

woah-pack-shot

### DIFF
--- a/tour-tip/component.js
+++ b/tour-tip/component.js
@@ -22,8 +22,8 @@ export default class TourTip extends Component {
 	renderCtas (data) {
 		if (data.content.ctas && data.content.ctas.length) {
 			return data.content.ctas.map(cta =>
-						<a href={cta.url} className="tour-tip__cta o-buttons o-buttons--standout" data-trackable="cta">{cta.label}</a>
-				);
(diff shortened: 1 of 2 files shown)


keirog:
For _some_ context, see [n-image](https://github.com/Financial-Times/n-image/blob/master/components/image.js#L69)

tour-tip/component.js
```diff
@@ -22,22 +22,29 @@ export default class TourTip extends Component {
...
+			);
 		}
 	}
 
 	renderImage (data) {
 		if (data.content.imageUrl) {
-			const isSvgExt = data.content.imageUrl.substr(-4,4) === '.svg';
-			const imageServiceOpts = (isSvgExt) ? {format: 'svg'} : {};
+			const imageAttrs = {
+				alt: data.content.imageAlt,
```
debugwand: looks good, might this ever be null? (in which case alt: data.content.imageAlt || '')

keirog: yep but [n-image](https://github.com/Financial-Times/n-image/blob/master/components/image.js#L38) has our back, which also [sets the `role` to `presentation` if so](https://github.com/Financial-Times/n-image/blob/master/components/image.js#L67)

PR #517, opened 2016-09-30
Welcome banner closeable version

debugwand:


welcome-message/template.html
```diff
@@ -8,17 +8,29 @@
...
 </div>
 
-<div {{#if @root.flags.nextFtTour}}data-o-component="o-expander" data-component="welcome-banner" data-o-expander-shrink-to="hidden" data-o-expander-toggle-selector=".n-welcome-banner__button--toggler" {{/if}}class="{{#if @root.flags.nextFtTour}}o-expander n-welcome-banner{{else}}n-welcome-message{{/if}} n-welcome-message--fixed" data-trackable="welcome" hidden>
-	{{#if @root.flags.nextFtTour}}
-		{{>n-ui/welcome-message/banner}}
-	{{else}}
+{{!-- supporting 3 different banner versions --}}
+
+
+{{#unless @root.flags.nextFtTour}}
```
keirog: Would it be ok if we switched it around to be `if`/`else`?

PR #523, opened 2016-09-30
feat(subscription-prompt): add dynamic pricing & discounts based on locale

keithamus:
Leaving this here as a 90% done "feature".

Will complete Monday

### DIFF
--- a/subscription-offer-prompt/index.js
+++ b/subscription-offer-prompt/index.js
@@ -1,9 +1,10 @@
 import SlidingPopup from 'n-sliding-popup';
 import {
-	noneTrue, resultExists, executeWhen,
+	noneTrue, result, resultExists, executeWhen,
 	getStorage, getSessionStorage, setStorage, getCookie,
 	dateInFuture, addToDate,
 	elementExists, createElement,
+	toCurrency
 } from './utils';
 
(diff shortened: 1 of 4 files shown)


subscription-offer-prompt/index.js
```diff
@@ -65,5 +66,23 @@ function createSubscriptionPrompt (html) {
 	};
 }
 
-export const initPrompt = executeWhen(createSubscriptionPrompt(createPopupHTML));
+function getSubscriptionPromptValues () {
+	const currencyCode = 'GBP';
```
keithamus: This needs to be configured to the user's currency code. I don't know how to get or derive this. 

@matthew-andrews @wheresrhys @ironsidevsquincy @ifyio advice?

wheresrhys: Don't think it's possible with ease at present. Once we ditch ie9 & 10 js support, and DAM gets deprecated, I would hope that membership provide us with a CORS enabled service for this sort of thing. As a stop gap you could add a json endpoint to product-selector to get this info

keithamus:
@wheresrhys updated code. Could you please re-review?

wheresrhys:
:+1:

PR #529, opened 2016-10-03
fixing nasty race condition with css loading

wheresrhys:
Not sure how long this has been a bug, but probably exacerbated by the fact that the code that initialises the mainCss promise now lives in a shared bundle and is therefore more likely to be cached than the page's css is, which means it loads too quickly for the inline `onload` handler, so having the onload handler just set a flag is no longer enough. I have a sneaking suspicion that possibly Chrome have tweaked the asynchronissity of the onload handler too (previously if the css file was cached it fired immediately, but now it appears to be asynchronous). No time to investigate fully though

The end result being that the mainCss promise often didn't resolve before. This fixes that

@phamann @ironsidevsquincy

### DIFF
--- a/js-setup/js/loader.js
+++ b/js-setup/js/loader.js
@@ -75,7 +75,6 @@ class JsSetup {
 					});
 			};
 		}
-
 		return Promise.resolve({
 			flags: flags,
 			appInfo: this.appInfo,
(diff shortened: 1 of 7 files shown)


ironsidevsquincy:
👍

PR #541, opened 2016-10-05
Add logic for anon data model to tour-tip component

phamann:
For the new myFT tour `/tour/myft` we need to show/hide tip call-to-actions based on whether or not the user is anonymous. This adds said support to the component rendering.

/cc @keirog

### DIFF
--- a/tour-tip/component.js
+++ b/tour-tip/component.js
@@ -20,11 +20,14 @@ export default class TourTip extends Component {
 	}
 
 	renderCtas (data) {
-		if (data.content.ctas && data.content.ctas.length) {
-			return data.content.ctas.map(cta =>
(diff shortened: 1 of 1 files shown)


tour-tip/component.js
```diff
@@ -20,11 +20,14 @@ export default class TourTip extends Component {
 	}
 
 	renderCtas (data) {
-		if (data.content.ctas && data.content.ctas.length) {
-			return data.content.ctas.map(cta =>
-				<a href={cta.url} className="tour-tip__cta o-buttons o-buttons--standout" data-trackable="cta">{cta.label}</a>
-			);
+		if((!data.content.ctas && !data.content.ctas.length) ||
+			(data.settings.isAnon && !data.settings.showCtaToAnon)) {
```
keirog: Any chance we could change to `hideCtaFromAnon`, for consistency?

PR #542, opened 2016-10-05
makes the css for the promo banner critical to prevent a loading-time…

pixelandpage:
Makes the css for the promo banner 'critical' to prevent a loading-time bug on slow devices. 

slack_for_ios_upload


myft-digest-promo/main.scss
```diff
@@ -1,3 +1,5 @@
+@include nUiMaybeCriticalStart('myftDigestPromo');
```
debugwand: there should be a matching nUiMaybeCriticalEnd

wheresrhys:
There might be a better way to achieve this, will come to talk it over tomorrow

myft-digest-promo/main.scss
pixelandpage: Good idea!

keirog:
I‘m confused by the absence of [`nUiApply`](https://github.com/Financial-Times/n-ui/blob/master/tour-tip/main.scss#L3) here, but I don‘t really understand how n-ui+Sass works 😕

pixelandpage:
Latest commit makes better use of nUiCritical and nUiApply. It also moves the promo init() JS function call to delay it until after the non-critical CSS has loaded.

myft-digest-promo/main.scss
```diff
@@ -1,108 +1,115 @@
...
+		float: none;
 		position: relative;
-		@include oGridRespondTo('M') {
-			@include oGridCenter;
-			@include oGridOffset(2);
+		background-color: getColor('warm-5');
+		border-bottom: 1px solid getColor('warm-3');
+		border-top: 1px solid getColor('warm-3');
+
+		&.n-myft-digest-promo--enabled {
```
keirog: We shouldn’t need to chain the classes here; should be ok to have the `.n-myft-digest-promo--enabled` as its own block, outside and under this `.n-myft-digest-promo` block

PR #543, opened 2016-10-05
Add shared tour-tip dismisser

keirog:
Front page:
fp

Article:
article

Stream page:
stream

### DIFF
--- a/tour-tip/lib/dismiss.js
+++ b/tour-tip/lib/dismiss.js
@@ -1,0 +1,74 @@
+const superstore = require('superstore-sync');
+
+let dismisserButtons = [];
+let tipContainerEls;
+let localStorageKey;
+
+function hasLocalStorage () {
+	const TEST_KEY = 'tour-tip-test';
+	const TEST_VAL = 'can-store';
+
+	superstore.local.set(TEST_KEY, TEST_VAL);
+	const retrieved = superstore.local.get(TEST_KEY);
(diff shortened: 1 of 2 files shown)


tour-tip/lib/dismiss.js
```diff
@@ -0,0 +1,74 @@
...
+	button.innerHTML = '<span class="o-buttons-icon__label">Dismiss tips on this page</span>';
+	button.className = 'tour-tip__dismiss-btn o-buttons-icon o-buttons-icon--icon-only';
+	button.setAttribute('data-trackable', 'dismiss');
+	button.setAttribute('title', 'Dismiss tips on this page');
+
+	return button;
+}
+
+function setUpTipDismisser (appendDismisserTo) {
+	const dismisserContainers = Array.from(document.querySelectorAll(appendDismisserTo));
```
debugwand: array from is poorly supported, https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from

keirog: We’re polyfilling it afaik, https://polyfill.io/v2/docs/features/#Array_from

keirog: Yep, we’re safe https://github.com/Financial-Times/n-polyfill-io/blob/master/lib/configurations.js#L4

PR #545, opened 2016-10-06
Sticky ad 4

VladDubrovskis:
**Changes:**

1. Removed all the right hand rail stickiness just to keep it simple
2. The sticky ads inits once the ad is being loaded on the page, that way we let it expand first. We could test if this works for now and see if it actually has got any extra view ability impact. And if its too low could improve it then?
3. IE is not ideal and sometimes still hover the menu a little but only for a split second.

Very similar to the original solution, just tidied up the event listeners and references

### DIFF
--- a/ads/js/sticky.js
+++ b/ads/js/sticky.js
@@ -94,7 +94,6 @@ Sticky.prototype.init = function () {
 		return;
 	};
 
-	window.addEventListener('scroll', this.eventScrollStart);
 	const fixedElementTopPosition = this.fixed.getBoundingClientRect().top;
 	this.fixed.style.zIndex = '23';
 	this.fixed.style.top = `${fixedElementTopPosition}px`;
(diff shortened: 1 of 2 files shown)


ads/js/sticky.js
```diff
@@ -1,111 +1,125 @@
...
 
 function Sticky (el, opts) {
-	if (!el) return;
-	this.el = el;
 	this.opts = opts || {};
+	this.fixed = el;
+	this.boundary = document.querySelector(this.opts.stickUntil);
+	this.eventdbScrollEnd = debounce(this.scrollEnd.bind(this), 300);
+	this.eventScrollStart = this.scrollStart.bind(this);
 	this.sibling = opts.sibling ? document.querySelector(opts.sibling) : null;
```
gvonkoss: this was specific to RHR, the stickey header technically needs a sibling. I think.

andrewgeorgiou1981:
👍

ads/js/sticky.js
```diff
@@ -1,111 +1,119 @@
...
-	} else if (!this.onScrollListener && this.el.offsetHeight >= 10) {
-		this.bindScroll();
+	});
+}
+
+Sticky.prototype.calculate = function () {
+	const scrollY = window.pageYOffset || window.scrollY;
+	const atBoundary = (scrollY - this.startScroll + this.fixedHeight) >= this.boundaryTop;
+	const isAbsolute = this.fixed.style.position === 'absolute';
+	const extraAbsolute = ((scrollY - this.extraHeight) >= 0);
```
adgad: Perhaps a more explanatory variable name here?

adgad:
I think (separate to this PR) we should add a class where we want it sticky, and only do it there. i.e. the myFT sponsorship probably shouldn't be sticky.

PR #552, opened 2016-10-07
fix the sticky behaviour when the device is rotated or window is resized

VladDubrovskis:
/cc @adgad @gvonkoss @andrewgeorgiou1981 

Forgot to add this in previous PR :(


ads/js/sticky.js
```diff
@@ -70,7 +70,7 @@ Sticky.prototype.scrollStart = function () {
 	// only do this work once
 	this.fixedHeight = this.fixed.offsetHeight;
 	this.startScroll = window.pageYOffset;
-	this.boundaryTop = this.boundary.getBoundingClientRect().top;
+	this.boundaryTop = this.boundary.getBoundingClientRect().bottom;
```
gvonkoss: maybe reconsider naming this boundaryTop

adgad:
👍

PR #553, opened 2016-10-10
pa11y setup

lc512k:
@leggsimon @aintgoin2goa `o-header-search-term` was a duplicate id which was breaking a11y tests. Would you mind looking over https://github.com/Financial-Times/n-ui/pull/553/files#diff-f0927248e9f4776d64eb2e13cce4e165R5 to see if this would have an effect elsewhere? I don't see this id being used in js code, but I'm not very familiar with the header.

@wheresrhys this covers the basics. Will look into adding other modules if necessary in a separate PR.

cc @pixelandpage this will help catch a11y errors that are common to all the user-facing apps.

### DIFF
--- a/.pa11yci.js
+++ b/.pa11yci.js
@@ -1,17 +1,38 @@
 module.exports = {
-	"defaults": {
-		"timeout": 1000,
-		"page": {
-			"viewport": {
-				"width": 320,
-				"height": 480
-			}
-		}
-	},
 	"urls": [
 		{
 			"url": "https://n-ui.ft.com/n-ui/test-page/" + process.env.CIRCLE_BUILD_NUM + "/test-page.html",
 			"timeout": 10000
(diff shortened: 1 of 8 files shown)


.pa11yci.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+	"defaults": {
+		"timeout": 1000,
+		"page": {
+			"viewport": {
+				"width": 320,
```
wheresrhys: Is it possible to run the test twice on different widths - some features are displayed very differently depending on the screen size

lc512k: I've added 4 configurations as https://github.com/Financial-Times/n-ui/pull/553/files#diff-ae14d879ae6b6a0f804e6be9ece8ba95R2

cc @rowanmanning

PR #579, opened 2016-10-24
only reset the values once cookie close event if the header advert is…

VladDubrovskis:
… still sticky

Also all of this code pretty much screams refactoring, but considering we only need it for next couple of weeks (this quarter?) not going to invest any more time in it


adgad:
I think sticky was intended for beyond this quarter 🙈 but 👍

VladDubrovskis:
Will create a story in backlog to clean up this and see and stick to end of tech debt we currently doing

PR #581, opened 2016-10-26
subnav title attribute added for a11y

pixelandpage:
To pass our pa11y test suite, button elements must have a title attribute.

### DIFF
--- a/header/partials/subnav/template-new.html
+++ b/header/partials/subnav/template-new.html
@@ -30,8 +30,8 @@
 
 				</div>
 			</div>
-			<button class="o-header__subnav-button o-header__subnav-button--left" aria-label="scroll left" aria-hidden="true" disabled>scroll left</button>
-			<button class="o-header__subnav-button o-header__subnav-button--right" aria-label="scroll right" aria-hidden="true" disabled>scroll right</button>
+			<button class="o-header__subnav-button o-header__subnav-button--left" title="scroll left" aria-label="scroll left" aria-hidden="true" disabled></button>
+			<button class="o-header__subnav-button o-header__subnav-button--right" title="scroll right" aria-label="scroll right" aria-hidden="true" disabled></button>
 		</div>
 	</div>
 </div>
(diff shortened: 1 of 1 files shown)


header/partials/subnav/template-new.html
```diff
@@ -30,8 +30,8 @@
 
 				</div>
 			</div>
-			<button class="o-header__subnav-button o-header__subnav-button--left" aria-hidden="true" disabled></button>
-			<button class="o-header__subnav-button o-header__subnav-button--right" aria-hidden="true" disabled></button>
+			<button title="subnav-spacing-btn-left" class="o-header__subnav-button o-header__subnav-button--left" aria-hidden="true" disabled></button>
```
debugwand: titles 99% of the time are problematic for screenreaders, let me dig up the docs to confirm

debugwand: this looks like the value belongs in the class? also use aria-label and 'scroll left' text

header/partials/subnav/template-new.html
```diff
@@ -30,8 +30,8 @@
 
 				</div>
 			</div>
-			<button class="o-header__subnav-button o-header__subnav-button--left" aria-hidden="true" disabled></button>
-			<button class="o-header__subnav-button o-header__subnav-button--right" aria-hidden="true" disabled></button>
+			<button class="o-header__subnav-button o-header__subnav-button--left" title="scroll left" aria-label="scroll left" aria-hidden="true" disabled></button>
```
debugwand: you can probably drop the title altogether

lc512k:
@debugwand Enforcing aria-label is something we plan to look into in the coming weeks. After we have a mechanism for that we could consider turning off in pa11y the item in the standard that checks for title, if we find that it's not useful anywhere.

For now, for apps to pass pa11y they need to pass every item in the standard. I would suggest leaving title in for now, so that my-ft-page can pass pa11y checks (even if a11y itself is not great yet--acceleration month coming up!) and we can have the error threshold removed. Otherwise, we are leaving the door open for more errors to silently creep in, negating some of the effort @pixelandpage has put into removing them these past few days.

PR #582, opened 2016-10-26
Disable sticky ads on ios

VladDubrovskis:
/cc @adgad @gvonkoss @andrewgeorgiou1981 

Making sure we do not init the sticky on iOS devices. Regex taken from apps team so should be reliable :)


andrewgeorgiou1981:
👍

PR #583, opened 2016-10-27
disable sticky ad on Android too

VladDubrovskis:


VladDubrovskis:
User agent detection is bad, but we have a story to revisit all of this code

PR #591, opened 2016-11-01
Handlebars version of the Tour Tip component

bjfletcher:


tour-tip/template.html
```diff
@@ -0,0 +1,27 @@
+{{#unlessAll data.settings.isAnon data.settings.hideFromAnon}}
```
keirog: Nice addition of `unlessAll`

PR #595, opened 2016-11-03
use o-forms-v3

JakeChampion:


JakeChampion:
Hmm, build failed because of a contrast issue on a visually hidden component. I'm not sure that this is an issue.

```
Errors in https://n-ui.ft.com/n-ui/test-page/2887/test-page.html:

• This element has insufficient contrast at this conformance level. Expected a
   contrast ratio of at least 4.5:1, but text in this element has a contrast
   ratio of 3.92:1. Recommendation: change background to #317f85.

   (#o-header-drawer > div > div:nth-child(3) > form > button > span)

   <span class="o-header__visually-hidden">Search</span>

PR #605, opened 2016-11-13
Myft hint updates

pixelandpage:
1. Fixes the anonymous user check to avoid using the broken `_locals` path. 
2. Hint will now only be shown if the user is not currently following anything in myFT. 
3. Font changed to more closely match the original design.
4. Some tidying up.


myft-hint/index.js
```diff
@@ -1,10 +1,18 @@
...
 
-function shouldShowHint (){
-	return getHintShownCount() < MAX_SHOWS;
+function shouldShowHint () {
+	// The hint should be shown if the user is not subscribed to any myFT topics
+	// and the MAX shows value hasn't been reached.
+	return myftClient.getAll('followed', 'concept')
+		.then(function (userConcepts) {
+			const noConceptsAdded = !(userConcepts.length > 0);
+			const showCountBelowMax = getHintShownCount() < MAX_SHOWS;
```
keirog: We could save quite a few hits to the API if we calculated the `showCountBelowMax` before/outside of the `myftClient.getAll` call

myft-hint/main.scss
```diff
@@ -54,7 +53,7 @@
 	}
 
 	.myft-hint__content {
-		@include oTypographyBody;
+		@include oFontsInclude(MetricWeb, $weight: semibold);
```
keirog: This doesn’t do what you’re probably going for. `oFontsInclude` only declares the name/url of the font, rather than tell `.myft-hint__content` to use it. It doesn’t affect the visuals of your hint at all, although the _removal_ of `oTypographyBody` does.

Options:
- if you’re happy with the visuals as is (I haven’t seen it), you can remove this line of code
- if you wanted it to be bold, you can replace this line with `@include oTypographySansBold('m')`, but that’ll knock your font-size up a touch ([for reference](https://github.com/Financial-Times/o-typography#sans))

Gimme a shout if I’m chatting nonsense, which seems likely

header/partials/header/top.html
```diff
@@ -26,7 +26,7 @@
 							<span class="o-header__visually-hidden">myFT</span>
 						</a>
 						{{#if @root.flags.myftHint}}
-							{{#unless @root.locals.anon.userIsAnonymous}}
+							{{#unless @root.anon.userIsAnonymous}}
```
keirog: This doesn’t look quite right… The anon object is [still in locals, no](https://github.com/Financial-Times/n-express/blob/master/src/anon/middleware.js#L11)?

myft-hint/index.js
pixelandpage: Excellent point! 👍

PR #612, opened 2016-11-17
Update page-attention time to 15s timeout

adgad:
Based on fact that it takes longer than 5s to read a page worth of content /cc @commuterjoy @adambraimbridge


adambraimbridge:
Note to self: I wonder if we could combine attention time, scroll depth, and any other measurements to calculate some indication of whether or not the user actually read the article or not. (like http://upworthy.github.io/2014/06/implementing-attention-minutes-part-1/)

PR #625, opened 2016-11-24
Rhys/typeahead update

wheresrhys:


header/js/suggestion-list.js
```diff
@@ -0,0 +1,15 @@
+const React = require('react');
+
+export class SuggestionList extends React.Component {
```
i-like-robots: can be stateless

header/js/typeahead-new.js
```diff
@@ -0,0 +1,221 @@
...
+			this.viewAllItemInnerHTML = '<button type="submit" data-trackable="view-all">View All Results</button>';
+		}
+
+		this.delegate = new Delegate(this.container);
+		this.bodyDelegate = new Delegate(document.body);
+		this.suggest = this.suggest.bind(this);
+		this.onType = debounce(this.onType, 150).bind(this);
+		this.onDownArrow = this.onDownArrow.bind(this);
+		this.onSuggestionKey = this.onSuggestionKey.bind(this);
+		this.onSuggestionClick = this.onSuggestionClick.bind(this);
```
i-like-robots: why not make this into a react component as well?

wheresrhys: I don't know react that well, and would refactoring bits of the component that don't actually need to repaint benefit the user anyway?

header/js/suggestion-list.js
i-like-robots:
```js
export (props) => {
}
```

header/js/suggestion-list.js
```diff
@@ -0,0 +1,15 @@
...
+	render () {
+		const suggestions = this.state.suggestions || [];
```
i-like-robots: Oh wait, you have state - this should be declared in the constructor

i-like-robots: But they'll be passed down right? So should be props?

PR #633, opened 2016-12-01
open survey in a new window

lc512k:
@i-like-robots Survey Monkey currently leaves the user in a blank page when they finish the survey :|

Ideally, they'd redirect them to the page they came from (could be any inside ft.com) but we've been told it's not possible. 

We've opted for opening the survey in a new window and then having survey monkey close itself when finished. Not ideal, but perhaps less disorienting than opening in the same window then offering a link back to ft.com when done (rather than the article they actually came from, for example)?


i-like-robots:
Sounds reasonable!

PR #634, opened 2016-12-01
Initial setup for pa11y testing components

GlynnPhillips:
This is my first draft for pa11y testing components, currently this will remove testing of the full demo page but that can easily be added back in I just wanted to keep it simple for now so it is easy to understand.

Some tidying up will be done if we agree on this approach.

Each component that we want to test will need a `pa11y-config.js` file in the root directory. This will respond with JSON that should also allow us to cover most scenarios in the future. For now it will allow us to flag the main template to load for testing and supply any data to the component that might be required.

Example contents for `pa11y-config.js`
```
{
    "entry": "template-name",
    "data": {} // any data required to render the component
}
```

All directories within the the root of `n-ui` are considered to be a component by default. This could be improved in the future with a file structure like the proposed by @wheresrhys in this branch https://github.com/Financial-Times/n-ui/tree/rhys/v3-structure

There is a `pa11yIgnore` array inside the `.pa11yci.js` file that can be used to remove any components that shouldn't be tested or for now to remove directories such as `node_modules`.

By assuming all directories are components by default means that CI and tests can fail when someone creates a new component and doesn't do the required pa11y set up. If they for whatever reason choose not to test a component they will have to physically opt out of pa11y testing and should be able to explain why in a pull request.

Components will be testable on the test-server via the route `/components/{component-name}`

/cc @lc512k @andygout @keirog


.pa11yci.js
```diff
@@ -1,41 +1,84 @@
...
-			url: url,
-			viewport: viewport
-		})
+			"url": `localhost:5005/components/${component}`
+		});
 	}
+});
+
+if(missingPa11yConfig.length) {
+	throw new Error(error(`Error: Components need to have a pa11y-config, components missing a pa11y-config: ${missingPa11yConfig.join()}`));
```
GlynnPhillips: I made this build fail purposefully to demo what will happen

https://circleci.com/gh/Financial-Times/n-ui/3072

I think it should also point to the readme which should explain how to set this up and in small print how to remove a component if really necessary

lc512k:
Looks great Glynn. Do you see some value in adding the full demo page test back in in the future or would it be completely redundant?

I like the idea of new components having to explicitly opt out of pa11y and having to justify

GlynnPhillips:
Right now I think there is still some value in the full demo page and as it is so simple to add in I don't see a reason not too.

There might be improvements we can do in the future to the components themselves what would result in the full demo having less value but all of this would need to be thought about more which we would not have time for now and would be creeping outside of the a11y remit

wheresrhys:
> By assuming all directories are components by default means that CI and tests can fail when someone creates a new component and doesn't do the required pa11y set up
:+1::+1:

_test-server/app.js
```diff
@@ -12,9 +13,9 @@ const app = module.exports = express({
 	hasHeadCss: true,
 	layoutsDir: path.join(process.cwd(), '/layout'),
 	viewsDirectory: '/_test-server/views',
+	partialsDirectory: process.cwd(),
```
wheresrhys: could you experiment with removing all the hacky stuff in `make serve` that copies a load of templates. This may fix that

_test-server/views/component-wrapper.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
```
wheresrhys: why not use `/layout/vanilla.html`

GlynnPhillips: The main reason was to be able to add a wrapper around the component but I agree would be nicer to use `vanilla` so I am in the process of switching that but it means `vanilla` will have a wrapper added based on a flag

PR #644, opened 2016-12-05
adding utils for working with cookies

aintgoin2goa:
@bjfletcher @phamann

### DIFF
--- a/utils/js/cookies.js
+++ b/utils/js/cookies.js
@@ -1,25 +1,20 @@
-const DEFAULT_OPTIONS = {
-	domain: location.hostname,
-	path: '/',
-};
-
 // Regex taken from https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie don't ask me how it works
 function getRegexForName (name){
 	return new RegExp(`(?:(?:^|.*;\\s*)${name}\\s*\=\\s*([^;]*).*$)|^.*$`);
 }
 
-function getOptions (opts){
-	const options = Object.assign({}, DEFAULT_OPTIONS, opts);
+function getOptions (options){
 	return Object.keys(options).map(key => {
+		let newKey = key;
 		if(key === 'maxAge'){
-			key = 'max-age';
+			newKey = 'max-age';
 		}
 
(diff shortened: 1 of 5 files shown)


utils/js/cookies.js
```diff
@@ -0,0 +1,39 @@
...
+
+function getOptions (options){
+	return Object.keys(options).map(key => {
+		let newKey = key;
+		if(key === 'maxAge'){
+			newKey = 'max-age';
+		}
+
+		if(options[key] instanceof Date){
+			options[key] = options[key] = options[key].toUTCString();
```
debugwand: `options[key] = options[key] = options[key].toUTCString();`  - why 2`options[key] = `

utils/js/cookies.js
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		return `${newKey}=${options[key]}`;
+	});
+}
+
+function set (name, value, opts){
+	const options = getOptions(opts).join('; ');
+	const cookie = `${encodeURIComponent(name)}=${encodeURIComponent(value)}; ${options}`;
```
aintgoin2goa: Cos cookies are supposed to be url encoded according to the spec.  Browsers handle it either way in most cases, but just trying to be correct

utils/js/cookies.js
aintgoin2goa: Cos I'm an idiot who types things twice types things twice

utils/js/cookies.js
ironsidevsquincy: Then we should un-encode them when `get`ting

ironsidevsquincy: And presumably also encode the name before `get`ting and `has`ing

aintgoin2goa: yeah good point

bjfletcher:
Great, thanks Paul!

Can we add a comment at the top of the utility:

https://github.com/WICG/async-cookies-api

that we can use as reference as potentially something we may want to use in the future.

(Note that we can't use that polyfill just yet because it uses `async`.)

PR #653, opened 2016-12-08
Hide the banner promo for anon users

tavvy:
Due to a bug in how the next-session api is reporting a users
products we are falesly showing the banner to SUB'd users.

This removes the logic that sets the isAnonUser flag until fixed.


commuterjoy:
+1

PR #672, opened 2016-12-13
Make the aria-label & title more descriptive for save buttons

keirog:
Addresses issue DAC-USER-SRJAWS-T10-01 of the accessibility report.

@lc512k @GlynnPhillips


lc512k:
@keirog what screen reader users might need is something like `save "Trump nominates Exxon chief Tillerson for US secretary of state" for later`, so that when they're navigating out of context they'll know what the button refers to:

screen shot 2016-12-13 at 21 46 38

keirog:
@lc512k Ah, I see. Will amend. Thanks!

PR #675, opened 2016-12-14
add viewStyle to hompage pageView context

aintgoin2goa:
@commuterjoy We talked about this ages ago - is to allow people to differentiate between "compact" and "standard" versions of the homepage

### DIFF
--- a/tracking/ft/index.js
+++ b/tracking/ft/index.js
@@ -3,7 +3,8 @@ const oGrid = require('o-grid');
 const oViewport = require('o-viewport');
 const nextEvents = require('./next-events');
 
-import {broadcast} from '../../utils'
+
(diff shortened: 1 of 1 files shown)


tracking/ft/index.js
```diff
@@ -77,6 +78,12 @@ const oTrackingWrapper = {
 				}
 			}
 
+			// if we're on the homepage add viewStyle = ("compact"|"standard") to allow people to differentiate
+			if(location.pathname === '/'){
+				const viewStyle = cookieStore.get('ft-homepage-view') || 'standard';
+				pageViewConf.context.viewStyle = viewStyle;
```
commuterjoy: Call it 'mode'

commuterjoy:
+1

PR #682, opened 2016-12-15
fix o-visual semver range

wheresrhys:
@alicebartlett @laurieboyes


alicebartlett:
Oh! I was suggesting we update this range in o-header and o-ads, but this works too!

wheresrhys:
rangify all the o-things

PR #698, opened 2017-01-06
upgraded preact

wheresrhys:
@i-like-robots 

If this passes ci ( a few preact components have ok test coverage) d'you think it's ok to release as a minor? Can always rollback if it goes boom


wheresrhys:
no big deal if not. I may - :scream: - do a major release towards the end of next week

i-like-robots:
This should be fine AFAIK

PR #699, opened 2017-01-08
Update .n-util-visually-hidden

lc512k:
Having `n-util-visually-hidden` with `height: 1px` causes screen readers to read out the hidden box before anything else in its parent in Chrome, which is not always what we want (see `Digital Accessibility Centre (DAC)` link in https://www.ft.com/accessibility using Voice Over).

Can you think of any issues removing `height: 1px` would cause visually @GlynnPhillips @i-like-robots ?

### DIFF
--- a/utils/main.scss
+++ b/utils/main.scss
@@ -34,12 +34,12 @@
 
 	.n-util-visually-hidden {
 		position: absolute;
-		clip: rect(0 0 0 0);
 		margin: -1px;
 		border: 0;
 		overflow: hidden;
 		padding: 0;
-		width: 1px;
+		width: 0;
+		height: 0;
 		white-space: nowrap;
 	}
 
--- a/utils/main.scss
+++ b/utils/main.scss
@@ -40,7 +40,6 @@
 		overflow: hidden;
 		padding: 0;
 		width: 1px;
-		height: 1px;
 		white-space: nowrap;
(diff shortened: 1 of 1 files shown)


i-like-robots:
This may be VO specific, JAWS reads the text in source order

GlynnPhillips:
I have had a play with this in a few of our more common user cases for `n-util-visually-hidden` and I  got an experience I would expect, so I think it might be worth playing devils advocate in this specific case.

When [snook](https://snook.ca/archives/html_and_css/hiding-content-for-accessibility) did the initial research for hiding elements the reason for `height: 1` was that VO would not read content inside elements with a height of 0. This was in 2011 and with more recently VO versions it would appear to be fine with `height: 0` and even better it also means it reads the elements in the correct order.

Considering that technique was created in 2011 and uses clip which is deprecated maybe someone should spend some time looking at if we can ditch clip for `width: 0; height: 0;` 😉 

I am not sure if this has any impact on other screen readers though or know when the beahviour snook saw in VO stopped happening.

lc512k:
@i-like-robots Yes, it's only VO and on Chrome, specifically. I haven't come across it elsewhere

@GlynnPhillips, nice one. `width: 0; height: 0;` without clip does the job! I'll update the PR

PR #701, opened 2017-01-10
Adding ticks for downloadable articles

aintgoin2goa:
@lc512k 
@tavvy 

If it finds articles that are marked as available to syndication it checks the users products (first in the cookie, or via session-next) to see if they are a syndication customer.  If they are a green tick is added, which is a link to download the article on the syndication platform.

screen shot 2017-01-10 at 17 36 42

### DIFF
--- a/.pa11yci.js
+++ b/.pa11yci.js
@@ -20,6 +20,7 @@ const config = {
 
 const pa11yIgnore = [
 	'.git',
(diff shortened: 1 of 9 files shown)


aintgoin2goa:
Still to do: cross-browser testing and adding the tick on the article page itself.

PR #706, opened 2017-01-12
Update pa11y testing to be more configurable

keirog:
This PR enables us to test any given component multiple times in pa11y-ci, each time with different pa11y settings (e.g. testing with different flags toggled).

Please check individual commits for marginally more context 😬 

@lc512k (@jkerr321 )


lc512k:
Brilliant! Thanks @keirog! 🎉

PR #709, opened 2017-01-12
deploying precompiled layots to s3

wheresrhys:
@kavanagh @i-like-robots @JakeChampion 

Jake - note, I will do the signing of the template in a separate PR - they're not consumed yet, and I don't want this PR to be a monster

### DIFF
--- a/_deploy/compile-layouts.js
+++ b/_deploy/compile-layouts.js
@@ -1,0 +1,36 @@
+const path = require('path');
+const fs = require('fs');
+const denodeify = require('denodeify');
+const readFile = denodeify(fs.readFile);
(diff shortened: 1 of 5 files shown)


_deploy/compile-layouts.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Handlebars = require('@financial-times/n-handlebars');
+
+Handlebars.standalone({
+	layoutsDir: path.join(process.cwd(), '/layout/'),
+	helpers: {},
+	directory: process.cwd()
+})
+	.then(hbs =>
+		shellpromise('mkdir ./dist/templates')
```
JakeChampion: could use the npm module `mkdirp` here

PR #725, opened 2017-01-19
Stop n-makefile from clobbering over .pa11yci.js

keirog:
Still not sure why it’s happening now though, and not before

@wheresrhys


keirog:
Oh I see… it’s the n-makefile upgrade

PR #735, opened 2017-01-25
hack to avoid negation of display:none on footer

wheresrhys:
@i-like-robots @adgad @gvonkoss

### DIFF
--- a/layout/main.scss
+++ b/layout/main.scss
@@ -50,9 +50,9 @@
 	}
 
 	@include nUiCriticalEnd('core');
-
+	// needed to stop css minimize merging the selectors above and below
 	// undo the above hiding of the footer
-	.n-layout__row--footer {
+	.n-layout__row--footer.n-layout__row--footer {
 		display: table-row;
 	}
 }


layout/main.scss
```diff
@@ -50,9 +50,9 @@
 	}
 
 	@include nUiCriticalEnd('core');
-
+	// needed to stop css minimize merging the selectors above and below
 	// undo the above hiding of the footer
-	.n-layout__row--footer {
+	.n-layout__row--footer.n-layout__row--footer {
```
gvonkoss: isn't it `.n-layout__row.n-layout__row--footer` ...? If the answer to that is now, then I'm confused 🙈

wheresrhys: This really is just a massive hack to stop our build converting 

`.n-layout__row-footer {display: none}
/*! head css comment */
.n-layout__row-footer {display: table-row}
```
into 

```
/*! head css comment */.n-layout__row-footer {display: none; display: table-row}
```

`.n-layout__row--footer.n-layout__row--footer` has the effect of having a different selector to `.n-layout__row--footer`, so the blocks won't get merged, but it doesn't actually change what gets targeted (though it does change the specicifity, which isn't ideal, but safe to do here... I think... as the classname isn't used anywhere else)

Not pretty, and need to find a better solution sometime, but for now it's a production bug

PR #741, opened 2017-01-27
Expose hasher to apps, and remove hashedAsset helper use

adgad:
@wheresrhys This look alright?

### DIFF
--- a/node/index.js
+++ b/node/index.js
@@ -98,7 +98,9 @@ module.exports = options => {
 		next();
 	});
 
-	let hasher = str => str;
+	let hasher = {
+		get: str => str
+	};
 	// verification that expected assets exist
 	if (options.withAssets) {
 		verifyAssetsExist.verify(app.locals);
@@ -113,6 +115,8 @@ module.exports = options => {
 			hasher
 		}));
 	}
+
+	app.hasher = hasher;
 
 	// Decorate responses with data about which assets the page needs
 	if (options.withAssets) {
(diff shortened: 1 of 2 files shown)


node/index.js
```diff
@@ -98,7 +98,9 @@ module.exports = options => {
 		next();
 	});
 
-	let hasher = str => str;
+	let hasher = {
```
adgad: wasn't sure about this and where it's used, but I assume it should have the same API as the actual hasher...

adgad:
And consumed here - https://github.com/Financial-Times/next-article/pull/1791

node/index.js
wheresrhys: Would you mind deleting https://github.com/Financial-Times/n-ui/blob/0caf851bce8265fbf12b00160c7d8b050398d944/node/index.js#L115.

Also I think it should be exposed on the app as just a function with a more descriptive name... `getHashedAssetUrl`?
(most of the code in the assets bits and pieces I imagine will be refactored massively as it's a mess, so it shouldn't really inform what the public api looks like)

wheresrhys:
Just realised this would completely unstyle ie8!!!

This conditional comment needs tweaking https://github.com/Financial-Times/n-ui/blob/26536bb4da0834c5d757b68f1935e549c92223f0/layout/partials/stylesheets.html#L1

adgad:
@wheresrhys I *think* this does the trick. Some of the styles look a bit naff on IE8 (cropping of the FT logo etc), I suspect we had those covered in main-ie8.css which no longer exists.

wheresrhys:
yeah -IE8 is technically no longer core experience, so we have the go ahead to be a bit broken

PR #742, opened 2017-01-27
Rename equities to securities

adgad:
@laurieboyes Doing this blind, does it look about right? (trying to rename the thing in the typeahead box


laurieboyes:
Ha! Yeah I think so 👍

PR #760, opened 2017-02-01
Amend Gift Article copy

mmockett:
Gift Article credits have increased from 10 to 20 for Premium users. Have amended the text to reflect this.

@laurieboyes are the short versions okay length-wise?


laurieboyes:
Yep, I've put them in context and they look fine 👍  I'll get this passing then merge and release

PR #762, opened 2017-02-01
Rhys/remove polling

wheresrhys:
@adgad Banishes the evil poller to bad idea hell. 

@andygnewman I've added a rebuild of user-facing apps after release. This unblocks work on n-ui, and there's some visibility of the state of apps here http://ombudsman.in.ft.com/repos?filter=nUi


Makefile
```diff
@@ -51,6 +51,8 @@ deploy: assets
 	node ./_deploy/compile-layouts
 	node ./_deploy/s3.js
 	$(MAKE) npm-publish
+	sleep 20
+	nht rebuild --all --serves user-page
```
adgad: I'm in two minds as to whether this is a good thing or not (but happy to try/see how it goes)

adgad:
LGTM

Makefile
wheresrhys: To avoid apps getting out of step we need some way of enforcing this. I figure this is better than slacking the repo owners to update

PR #768, opened 2017-02-06
Remove: Front Page Fast FT Journey flag reference

andygout:
Given the flag was removed on 31 Jan 2017.

screen shot 2017-02-06 at 11 09 32


adgad:
Good spot Mr Gout! I don't think v2 was a thing when I deleted the flag :)

PR #769, opened 2017-02-06
update o-buttons

gvonkoss:
@wheresrhys  It would be nice to use the new versions of the buttons on the topper — the biggest change, I believe, are the sharp corners, which AFAIK are meant to be incorporated into the whole site anyway.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -26,7 +26,7 @@
     "n-sliding-popup": "^1.0.0",
     "next-myft-client": "^6.4.2",
     "o-ads": "^6.0.0",
-    "o-buttons": "4.4.4",
+    "o-buttons": "^4.4.4",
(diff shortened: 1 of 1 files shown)


bower.json
```diff
@@ -26,7 +26,7 @@
     "n-sliding-popup": "^1.0.0",
     "next-myft-client": "^6.4.2",
     "o-ads": "^6.0.0",
-    "o-buttons": "4.4.4",
+    "o-buttons": "^4.4.4",
```
kavanagh: Wouldn't we need to force the issue to make next-article work? i.e. `o-buttons": "^4.5.0`

kavanagh:
LGTM :shipit:

PR #772, opened 2017-02-06
Ads new targeting parameter

andrewgeorgiou1981:
@andygnewman


ads/js/utils.js
```diff
@@ -31,6 +31,23 @@ function getLayoutName () {
...
+function getVariantNames () {
+	let abState = '-';
+	const abStateEl = document.querySelector('[data-ab-state]');
+	if(abStateEl) {
+		abState = abStateEl.getAttribute('data-ab-state');
+	}
+	return abState;
+}
+
+let consolidateMetrics = function (container, metrics){
```
andygnewman: why use `let` rather than just `function consolidateMetrics () {}`

ads/js/utils.js
```diff
@@ -31,6 +31,23 @@ function getLayoutName () {
...
+	let slotName = container.dataset['oAdsName'];
```
andygnewman: should this be a `const` rather than a `let`?

ads/js/utils.js
```diff
@@ -31,6 +31,23 @@ function getLayoutName () {
 	return layout;
 }
 
+function getVariantNames () {
```
andygnewman: Perhaps would be clearer to call this function getABTestState?

ads/js/utils.js
andrewgeorgiou1981: you're right, this can just be a function definition, rather than assign it to a var. i'll update

ads/index.js
```diff
@@ -28,6 +28,15 @@ function initOAds (flags, appName, adOptions) {
 		const containers = [].slice.call(document.querySelectorAll('.o-ads'));
 		slotCount = containers.length;
 		utils.log.info(slotCount + ' ad slots found on page');
+		if (flags && flags.get('AdsMetricsInOneKey')) {
+			let metrics = '';
+			metrics += 'adUnit=' + initObj.gpt.site + '/' + initObj.gpt.zone;
+			metrics += (res.targeting.get().pt) ? '|pageType=' + res.targeting.get().pt : '';
+			metrics += (res.targeting.get().res) ? '|res=' + res.targeting.get().res : '';
```
andygnewman: add in the mvt value

PR #789, opened 2017-02-08
Rhys/out of hours

wheresrhys:


wheresrhys:
Yeah I'm not sure where to put the threshold. The advantage of having it wide is it pushes code changes out to all our apps quickly so we get to find problems quickly, rather than stacking up every change from 4pm till 11am and then suddenly a load of features get released at once at 11am

We shall see how it goes

PR #793, opened 2017-02-10
don't show compact view prompt in drawer to anons

TheoLeanse:
@aintgoin2goa @adgad 

not sure why there are two versions but hey


adgad:
👍

TheoLeanse:
also fyi @wheresrhys I'm adding / fixing tests

PR #806, opened 2017-02-14
Noob friendly

adambraimbridge:
image

~~ and ~~

image

~~ and ~~

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,8 @@ a11y: test-build pally-conf
 	cp -rf $(shell cat _test-server/template-copy-list.txt) bower_components/n-ui
 	PA11Y=true node _test-server/app
 
-# TODO: Figure this out: Because `run` never ends, `nightwatch` and `a11y` will never execute.
-# Unless ... are they execured in different processes? how does that work? Where's the output?
+# Note: `run` executes `node _test-server/app`, which fires up exchange, then deploys 
+# a test static site to s3, then exits, freeing the process to execute `nightwatch a11y`.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -42,8 +46,20 @@ a11y: test-build pally-conf
 	cp -rf $(shell cat _test-server/template-copy-list.txt) bower_components/n-ui
 	PA11Y=true node _test-server/app
 
-test: verify pally-conf test-server test-unit test-build run nightwatch a11y
+# TODO: Figure this out: Because `run` never ends, `nightwatch` and `a11y` will never execute.
+# Unless ... are they execured in different processes? how does that work? Where's the output?
+test: developer-note verify pally-conf test-server test-unit test-build run nightwatch a11y
```
adambraimbridge: @wheresrhys: Let's chat about this when you have time and inclination. Learning opportunity for me.

Previously, it was `test: verify test-unit test-build serve nightwatch a11y`, and `serve` changed to `run`, here:  
https://github.com/Financial-Times/n-ui/pull/659/files#diff-b67911656ef5d18c4ae36cb6741b7965R37

... but [even then](https://github.com/Financial-Times/n-ui/blame/e2a7d4d7546d687da72939ad77b1bfc5894aa46c/Makefile#L18), `serve` ran `node _test-server/app`, so I don't see how `nightwatch a11y` gets executed. If they do, I still don't understand how that works, or where's the output?

adambraimbridge: Aha! I figured it out with @adgad — in CircleCI, `make run` exits when it is done, freeing the process so it can execute ` nightwatch a11y`.

PR #807, opened 2017-02-15
Rhys/simplify deploy script

wheresrhys:
We no longer need to deploy e.g. v2, v2.1 and v2.2.2 as n-ui triggers builds of every app when a version is released


wheresrhys:
confirmed it still deploys ok https://www.ft.com/__assets/n-ui/cached/rhys-test8/es5.min.js

PR #811, opened 2017-02-16
colour change to Powr of you orange, and copy update

jkerr321:
@lc512k


onishiweb:
@jkerr321 can you add an issue to `o-colors` about this color update, we're working on `o-colors` at the moment and this won't have been caught in our audit. Could you add a reason the orange needed to be changed too please :)

PR #814, opened 2017-02-20
n-ui 4 candidate

wheresrhys:
If you only review one n-ui PR, please make it this one

Changelog
- `n-ui/env` (sass) and `n-ui/utils` (js) have now moved, largely unaltered aside from the top level name, into `n-ui-foundations` so we can avoid circular dependencies when components outside of n-ui want to use primitive features
- js aliases for origami components are gone i.e. `require('n-ui/expander')` will no longer work and should be replaced by `require('o-expander')`
- A few non-primitive components - o-video, o-expander, o-overlay - are no longer bundled in the shared js bundle
- `nUiConfigure()` mixin no longer exists. The main entry point for the module is `main.scss`. Critical path css is to be achieved by using the `nUiStylesheetStart()` and `nUiStylesheetEnd()` mixins. subcomponents whose styles are not bundled by default should be included using normal sass `@import`s e.g `@import "n-ui/tour-tip"`
- also need to merge, release major version and update to this https://github.com/Financial-Times/n-express/pull/451
- webpack 2?


header/main.scss
```diff
@@ -2,127 +2,126 @@
...
-		width: 100%;
-		box-sizing: border-box;
-		border-top: 1px solid #cec6b9;
-		border-bottom: 1px solid #cec6b9;
-
-		.o-header__viewtoggle__label {
-			padding: 12px 12px 12px 0;
-		}
+	.o-header__mega {
+		@include nUiZIndexFor('meganav');
```
leggsimon: Wasn't `nUiZIndexFor` deleted in v3?! How does this even work?

header/main.scss
```diff
@@ -2,127 +2,126 @@
...
+	.o-header--sticky {
+		display: none;
 	}
 
-	@media print {
-		.o-header--sticky {
-			display: none;
+	// skip links for keyboard users
+	.n-skip-link {
+		@include nUiZIndexFor('notification');
```
leggsimon: Hmmmm 🤔

header/main.scss
```diff
@@ -2,127 +2,126 @@
...
-	.o-header__drawer-homepage {
-		border-top: 2px solid getColor('pink-tint2');
+@include nUiStylesheetEnd('head-n-ui-core');
+
+@include oHeaderDrawer;
+@include oHeaderSticky;
+@include nTypeahead;
+// fix up z-indexes
+.o-header__drawer {
+	@include nUiZIndexFor('drawer');
```
leggsimon: Ok, I guess not

header/main.scss
```diff
@@ -2,127 +2,126 @@
...
+}
+.o-header--sticky {
+	@include nUiZIndexFor('sticky-header');
```
leggsimon: Yeah, almost certainly not

syndication/index.js
```diff
@@ -1,4 +1,4 @@
-import {$$, $} from '../utils';
+import {$$, $} from 'n-ui-foundations';
```
leggsimon: Ummmmm, what is `$` and `$$`??

typeahead/main.scss
```diff
@@ -6,7 +6,7 @@
 
 	.n-typeahead {
 
-		@include nUtilZIndexFor('typeahead');
+		@include nUiZIndexFor('typeahead');
```
leggsimon: Fine! I was wrong!

header/main.scss
wheresrhys: In discussions with matt h we thought it'd be a good idea to allow for a small family of `n-ui` things (certainly having the lowest level utilities in a separate component is useful), but we don't particularly want long namespaces, so letting anything in an `n-ui-*` component use the `nUi` namespace

So the short answer is, it's in the `n-ui-foundations` component

syndication/index.js
wheresrhys: document.querySelector and document.querySelectorAll helpers. Always been there, but not widely used.

ads/js/metrics.js
```diff
@@ -1,4 +1,4 @@
-import { broadcast } from '../../utils'
+import { broadcast } from 'n-ui-foundations'
```
i-like-robots: Is this module just for compatibility?

wheresrhys: which one - ads/js/metrics.js or n-ui-foundations?

i-like-robots:
initially when n-ui was created the the "test" before adding new components to it was _is this component used in 3 or more apps?_. This may not have been honoured but it was straightforward on face value.

Is there a similar 'razor' for deciding whether code lives in this repo or n-ui-foundation?

ads/js/metrics.js
i-like-robots: The former... or, why is this not in the n-ui-foundation repo? Have added a comment below

wheresrhys:
foundation = may be used within components/component demos

adambraimbridge:
Re: The final bullet point in the description: "webpack 2?"

There is a PR for updating webpack to version 2, here in n-webpack: https://github.com/Financial-Times/n-webpack/pull/57/files

^ That's not going to have much affect, and might start breaking app builds, because they have their own version of webpack (and other build tools like babel). Also, they include `n-ui`, which also has another version of webpack + build tools. 

So the idea is to move all the build tools to `n-webpack`, so that when we need to update versions, we only need to do it in one place. 

I've been working towards that goal, and removed webpack + build tools from https://github.com/Financial-Times/next-feedback/pull/68 and https://github.com/Financial-Times/next-article/pull/1853 

So now the question is, should we remove webpack from `n-ui` first, or from all the repos that require `n-ui`? 

https://github.com/search?l=&p=1&q=org%3AFinancial-Times+n-ui+extension%3Ajson&ref=advsearch&type=Code&utf8=%E2%9C%93

PR #837, opened 2017-02-24
tryout polyfill bundle on same domain

wheresrhys:


node/lib/assets.js
```diff
@@ -93,7 +93,17 @@ function init (options, directory, locals) {
 				res.locals.nUiConfig = nUiConfig;
 
 				// work out which assets will be required by the page
-				const polyfillRoot = `//${flags.polyfillQA ? 'qa.polyfill.io' : 'next-geebee.ft.com/polyfill'}/v2/polyfill.min.js`;
+
+				let polyfillRoot;
+
+				if (flags.polyfillQA) {
+					polyfillRoot = '//qa.polyfill.io/v2/polyfill.min.js';
```
JakeChampion: Why not force https?

wheresrhys: occasional locally we need to be on http

JakeChampion: Even for third-party resources?

PR #841, opened 2017-02-28
Rhys/await multiple css

wheresrhys:
Reviews most welcome. 3 major changes here

 1. now when a lazy loaded stylesheet loads it queries for the existence of any other styles are waiting to load, and only fires the load event if they all have.
 2. Introduces an idea of how apps can specify multiple stylesheets per page
```
				res.locals.cssBundles = {
					inline: [],
					lazy: [],
					blocking: []
				};

```
These are set to sensible defaults (i.e. what all our apps aside from stream-page currently have), but can be overwritten by your controller. This will be a breaking change as it was too tricky maintaining support for the existing stylesheet variant implementation
3. removed the `withHeadCss` option. It's influence has drifted over time to mean 'use shared n-ui bits and pieces'... and all apps are doing that now, so n-ui can enforce a strong opinion about this I reckon

TODO
- [x] as it's a breaking change may as well change the promise name to `allStylesLoaded`
- [x] as templates come from the npm installed version can add a build step to convert the css-loader to its inline template version


node/lib/assets.js
```diff
@@ -112,36 +122,27 @@ function init (options, directory, locals) {
...
-					// define which css to output in the critical path
-					if (options.withHeadCss) {
-						// variants of head-n-ui-core no longer exist, but the app may not necessarily
-						// have successfully fetched head-n-ui-core from network, so fallback to the variant
-						// file while trying it out
-						if ('head-n-ui-core' in stylesheets) {
-							res.locals.criticalCss.push(stylesheets['head-n-ui-core'])
-						} else if (`head${cssVariant}-n-ui-core` in stylesheets) {
-							res.locals.criticalCss.push(stylesheets[`head${cssVariant}-n-ui-core`])
+					res.locals.cssBundles.inline = res.locals.cssBundles.inline.reduce((str, name) => {
```
wheresrhys: should memoize this

node/lib/assets.js
```diff
@@ -112,36 +122,27 @@ function init (options, directory, locals) {
...
-					}
+						return str + stylesheets[name];
+					}, '');
 
-					res.locals.cssBundles.push({
-						path: hasher.get(`main${cssVariant}.css`),
-						isMain: true,
-						isLazy: options.withHeadCss
-					});
+					res.locals.cssBundles.lazy = res.locals.cssBundles.lazy.map(getStylesheetPath);
```
wheresrhys: all these things running over both sets of styles... should abstract out into a function

wheresrhys:
Closing in favour of a pull into v5 branch

PR #842, opened 2017-02-28
Add: Counter ad blocking feature

andygnewman:
cc: @andrewgeorgiou1981 

- Replaces existing ad-blocking check (that existed in tracking/third-party) and expands to 100% of users with a spoor id

- Includes nCounterAdblocking component (https://github.com/Financial-Times/n-counter-ad-blocking)

- See the README in nCounterAdBlocking for details on how that component works (https://github.com/Financial-Times/n-counter-ad-blocking/blob/master/README.md)

To Do
- [x] Confirm with Ammit team that they are happy for the 100% ad-block check before merging


tom-parker:
👍

PR #843, opened 2017-02-28
first stab at an n-ui build tool

wheresrhys:


leggsimon:
What problem does this solve?

leggsimon:
so it'll make our local builds much faster?

wheresrhys:
that's one of the perks. 

reflecting on it, the main reason is to encapsulate all of n-ui's business in n-ui. No worrying about needing to coordinate some change with a related change in a build tool. If it's all in one repo then it's easier to reason about and manage

wheresrhys:
Well, it's an opt-in, so just gonna merge. If it a hinderance people can scream at me

PR #848, opened 2017-03-02
Remove webpack

adambraimbridge:
This is passing `make test-dev` locally. 

In next-article, I did the `make clean install build-production` test against the `financial-times/n-ui#remove-webpack` branch, and it passed.

image


adambraimbridge:
@wheresrhys This PR not only removes webpack & related packages, but also some packages that were not required anywhere — at least, as far as I can detect.

package.json
```diff
@@ -59,29 +44,24 @@
...
     "pa11y-ci": "^0.3.0",
-    "postcss-loader": "^0.9.1",
     "proxyquire": "^1.7.10",
-    "sass-loader": "^3.2.0",
     "semver": "^5.3.0",
     "sinon": "^1.17.4",
     "sinon-chai": "^2.8.0",
     "supertest": "^2.0.1",
-    "webpack": "^1.13.1"
+    "@financial-times/n-webpack": "^2.0.0"
```
wheresrhys: for now needs to still be a prod dependency. Even though it's a dev tool, when n-ui is installed by an app it will only install its dependencies, not devDependencies

package.json
```diff
@@ -94,7 +74,6 @@
     "preact-compat": "^3.11.0",
     "promise-rat-race": "^1.4.0",
     "proptypes": "^0.14.3",
-    "semver": "^5.3.0",
```
wheresrhys: I have a feeling this is needed in production too, in order for the app to decide which version of n-ui css to pull from the network

PR #852, opened 2017-03-06
verify that assets get built before deploy

wheresrhys:


adambraimbridge:
Odd that coveralls is passing, but seems to be causing circleci to fail...

PR #862, opened 2017-03-08
updated offerId to annual rather than monthly as supplied by marketing

jkerr321:


jkerr321:
Merging with agreement from @tavvy. I've checked prices against new offerId and they match

PR #870, opened 2017-03-13
Rhys/fix expander error

wheresrhys:


keirog:
@wheresrhys You’ll have to bring a couple of bits of this back to life, as it breaks functionality of the compact view feature. The `--fixed` version of the panel is needed for when all these conditions (iirc) are true (which was fun to figure out last week):
- `compactView` MVT flag is set to `Active`
- localStorage prop `n-welcome-message-ad-view-count` is < 3 (I tested with 0)
- localStorage prop `n-welcome-message-compact-ad-collapsed` is _removed_ if it exists
- you’re not already using the compact view (set via the menu drawer)

It’s a sticky panel that advertises the compact view feature. Gimme a shout in the morning if that’s not clear. Moshe owns this if that helps (and he will notice its disappearance—I did a similar thing last week, hence me finding out the above)

PR #872, opened 2017-03-14
WIP Has tabbed

lc512k:
@keirog is this sort of what you had in mind?

### DIFF
--- a/tabbing/index.js
+++ b/tabbing/index.js
@@ -1,0 +1,12 @@
+module.exports = {
+	init: function (flags) {
+		if (flags.get('tabe')) {
+			window.addEventListener('a11yKeyboardFocus', function (e) {
+				if (e.keyCode === 9) {
+					// don't do this if element is an input
+					// find element with js class in it, add has-tabbed class to it
+				}
+			});
+		}
+	}
+};
--- a/tabbing/index.js
+++ b/tabbing/index.js
@@ -1,10 +1,18 @@
 module.exports = {
 	init: function (flags) {
-		if (flags.get('tabe')) {
-			window.addEventListener('a11yKeyboardFocus', function (e) {
-				if (e.keyCode === 9) {
-					// don't do this if element is an input
-					// find element with js class in it, add has-tabbed class to it
+
+		if (flags.get('a11yKeyboardFocus')) {
+			const container = document.querySelector('.js');
+
+			window.addEventListener('keyup', function (e) {
+				const pressedTab = e.keyCode === 9;
+				const hasHasTabbedClass = container.className.indexOf('has-tabbed') !== -1; // needed for older IE vs classList?
+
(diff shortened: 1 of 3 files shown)


keirog:
Ah, kind-of; I’ll comment in the diff

tabbing/index.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+	init: function (flags) {
+
+		if (flags.get('a11yKeyboardFocus')) {
+			const container = document.querySelector('.js');
```
keirog: As far as checking for JS folks goes, I was thinking of just handling that [in the CSS](https://github.com/Financial-Times/n-ui-foundations/blob/9e66637b170f83755c93bbdb8064ce7854ac30d4/util/main.scss#L67), where 
```
.has-tabbed {
```
would become
```
.no-js,
.has-tabbed {
```

tabbing/index.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
keirog: I wonder if this script should live [inline in the head](https://github.com/Financial-Times/n-ui/blob/7fbadc0e8a4cff719b961fba924a7f4f0e4ad662/layout/wrapper.html#L64) rather than be loaded externally. Then JS people on slow/flaky connections won’t get a Moment of Rubbishly-Focussed Links (MORFL 😂 )

tabbing/index.js
```diff
@@ -0,0 +1,20 @@
...
+	init: function (flags) {
+
+		if (flags.get('a11yKeyboardFocus')) {
+			const container = document.querySelector('.js');
+
+			window.addEventListener('keyup', function (e) {
```
keirog: If we did end up inlining this code we’d have to handle `attachEvent/detachEvent`, as well as avoiding ES6 bits as you mentioned offline

tabbing/index.js
```diff
@@ -0,0 +1,20 @@
...
+				const pressedTab = e.keyCode === 9;
+				const hasHasTabbedClass = container.className.indexOf('has-tabbed') !== -1; // needed for older IE vs classList?
+
+				// Don't activate focus style when tab pressed inside a form elements
+				const activatesFocusStyle = !e.target.type;
```
keirog: Nice

tabbing/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+				if (pressedTab && !hasHasTabbedClass && activatesFocusStyle) {
+					container.className = container.className + ' has-tabbed';
+				}
+			});
```
keirog: Maybe `removeEventListener` after running?

keirog:
Sorry for the wall of comments!

lc512k:
This is great, thanks for these!!! 🙌

PR #877, opened 2017-03-16
new entry points for subcomponents

wheresrhys:
@adgad @gvonkoss


adgad:
👍

_entry.js
```diff
@@ -7,6 +7,9 @@ module.exports = function (withPreact, exclusions) {
 		'n-ui/ads': 'window.ftNextUi._ads',
```
adgad: Do these still need to exist?

wheresrhys: technically it'd be a breaking change to remove them.

I really wish I'd just exposed tracking, ads and typeahead as properties on the main export, but didn't think to

PR #880, opened 2017-03-16
correct measurement of head css size

wheresrhys:


server/lib/stylesheet-manager.js
```diff
@@ -3,11 +3,51 @@ const denodeify = require('denodeify');
...
+				metrics.histogram('head_css_size.gzip', gzip);
+				metrics.histogram(`head_css_size.gzip.${hash}`, gzip);
+			})
+
+
+		return concatenatedStylesCache[hash];
+	},
+
+	init: (options, directory) => {
 		const headCsses = fs.readdirSync(`${directory}/public`)
```
adgad: Is this no longer _head_ cssess then? That filter would also pick up non-head CSSes?

PR #886, opened 2017-03-22
[FIX] [pa11y] sticky nav for anon users

Lily2point0:
Fixes:
- <a> tags with no content or href
- missing 'Subscribe' and 'Sign in' buttons for anonymous users


andygout:
Not entirely sure what is meant by the first point here `Fixes: tags with no content or href` but feel it's worth merging this and getting the fix out while we work on a test.

andygout:
Oh, I see - it fixes the pa11y tests.

PR #892, opened 2017-03-28
Amend: enable tooltip to be used on non-block elements

andygnewman:
offsetLeft gives incorrect values for non-block elements

I have a usecase for putting this on a list item so have refactored it to use clientRects() to get the correct left positioning.

Also have ensured that the arrow is centred rather than offset slightly.

### DIFF
--- a/components/n-ui/tooltip/index.js
+++ b/components/n-ui/tooltip/index.js
@@ -79,7 +79,7 @@ export class NToolTip{
 
 	position (){
 		const elRect = this.el.getClientRects()[0];
-		const parentRect = this.el.parentNode.getClientRects()[0];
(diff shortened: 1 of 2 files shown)


components/n-ui/tooltip/index.js
```diff
@@ -78,7 +78,9 @@ export class NToolTip{
 	}
 
 	position (){
-		let left = (this.el.offsetLeft - (this.tooltip.offsetWidth / 2));
+		const elRect = this.el.getClientRects()[0];
+		const parentRect = this.el.parentNode.getClientRects()[0];
```
i-like-robots: I'm assuming this element should be `position: relative` but in case it isn't, should this be `offsetParent`?

andygnewman:
Thanks have swapped it over.

PR #894, opened 2017-03-29
Apply: o-header for sticky header anon user links

andygout:
`Sign In` and `Subscribe` links should behave as they do on `o-header` demo (confirmed with Sue Vago): https://www.ft.com/__origami/service/build/v2/demos/o-header@6.14.1/sticky-header

These changes give those links the same styling and show/hide logic.


i-like-robots:
🚢

PR #897, opened 2017-03-31
[FIX] added tab focus on offer popup

Lily2point0:
- adds tab focus when popup open
- resumes focus when popup dismissed, set `tabindex` to `-1`
- option to dismiss pressing `ESC`

### DIFF
--- a/components/n-ui/subscription-offer-prompt/lionel.js
+++ b/components/n-ui/subscription-offer-prompt/lionel.js
@@ -74,12 +74,56 @@ const createPopupHTML = values =>
 
 
 const createSubscriptionPrompt = values => {
+	let focusedElementBeforePrompt;
+	let focusableElementsStrings = ['.subscription-prompt--subscribe-btn', '.n-sliding-popup-close'];
+
 	const subscriptionPrompt = createPopupHTML(values);
-	subscriptionPrompt.onClose = setPromptLastClosed;
+	let focusableElements = subscriptionPrompt.querySelectorAll(focusableElementsStrings);
+	focusableElements = Array.prototype.slice.call(focusableElements);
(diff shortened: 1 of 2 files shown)


components/n-ui/subscription-offer-prompt/lionel.js
```diff
@@ -74,12 +74,56 @@ const createPopupHTML = values =>
...
+				}
+			} else {
+				if(document.activeElement === lastTabStop) {
+					e.preventDefault();
+					firstTabStop.focus();
+				}
+			}
+		}
+
+		if(e.keyCode === 27) {
```
lc512k: can we have a line of doc for what key maps to 27 please? Same for 9 above. Thanks!

components/n-ui/subscription-offer-prompt/lionel.js
```diff
@@ -74,12 +74,56 @@ const createPopupHTML = values =>
...
+
 	const subscriptionPrompt = createPopupHTML(values);
-	subscriptionPrompt.onClose = setPromptLastClosed;
+	let focusableElements = subscriptionPrompt.querySelectorAll(focusableElementsStrings);
+	focusableElements = Array.prototype.slice.call(focusableElements);
+
+	subscriptionPrompt.onClose = () => {
+		setPromptLastClosed();
+
+		if(focusedElementBeforePrompt !== undefined)	focusedElementBeforePrompt.focus();
```
lc512k: Ideally we'd have `focusedElementBeforePrompt.focus();` in a new line, enclosed in `{}`s even if it's one line, to keep with Next standards(ish) - My bad, I didn't mention that before! 🙈

PR #898, opened 2017-03-31
Amend: swap out h3 for div in search results heading

andygnewman:
cc: @lc512k 

The `<h3>` in the typeahead search results conflicts with the heading hierarchy of the page.

All styling is based on the class name, so replacement with a `<div>` should have no presentational impact.


andygnewman:
Was covered in https://github.com/Financial-Times/n-ui/commit/8c97d8fc8ce08a46e8327719d8e0334590dadce8

PR #902, opened 2017-04-05
Headings

lc512k:


leggsimon:
Any update on this @lc512k  ?

lc512k:
@wheresrhys 
Sorry, I missed this. 

Without this there's a risk of introducing a headings error via n-ui that then will show up in other apps (that do check for this rule).

PR #909, opened 2017-04-07
Remove default welcome banner

i-like-robots:
With permission from Moshe!


keirog:
Oh just saw this, yayyy 🎉

PR #927, opened 2017-05-03
don't handle anonymity within the template

leggsimon:
So as far as I can tell, currently, the only place this handle bars template is used is within the stream-page app. Where the use is they already check if the user isn't anonymous so I think this doesn't need to have the check in here and rather consumers of the template should decide what to do with the template.


adgad:
seems to make sense 👍

PR #929, opened 2017-05-03
remove complicated anonymous logic from template

leggsimon:
Removes logic that should really exist outside of the template as to whether you, as the consumer, wants to render the CTAs. If not, remove the CTAs property from the model.

### DIFF
--- a/components/n-ui/tour-tip/template.html
+++ b/components/n-ui/tour-tip/template.html
@@ -10,19 +10,11 @@
 			{{#if data.content.body}}
 				<p class="tour-tip__body">{{{data.content.body}}}</p>
 			{{/if}}
-			{{#ifAll data.content.ctas data.content.ctas.length}}
-				{{#if data.settings.isAnon}}
-					{{#unless data.settings.hideCtaFromAnon}}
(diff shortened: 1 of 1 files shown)


components/n-ui/tour-tip/template.html
```diff
@@ -10,19 +10,11 @@
...
-							<a href="{{{url}}}" class="tour-tip__cta o-buttons o-buttons--standout" data-trackable="cta">{{label}}</a>
-						{{/each}}
-					{{/unless}}
-				{{else}}
-					{{#each data.content.ctas}}
-						<a href="{{{url}}}" class="tour-tip__cta o-buttons o-buttons--standout" data-trackable="cta">{{label}}</a>
-					{{/each}}
-				{{/if}}
-			{{/ifAll}}
+			{{#if data.content.ctas}}
```
GlynnPhillips: Is `if` needed?

leggsimon: wasn't sure whether to go for if anon remove cta property altogether or to make it an empty array. Thought the easiest would be just to remove the property

PR #932, opened 2017-05-04
make tracking and ads resilient to changing concept id to v2

wheresrhys:
So the idea is that we add a data-concept-idv1 attribute which duplicates the current data-concept-id value for now. Then when we update streams to v2 we change only the value in data-concept-id. Once ads and tracking stacks are fully updated to be able to use v2 we stop referencing the v1 property here and can delete the property


tom-parker:
👍

PR #938, opened 2017-05-17
Update theme-tag

TheoLeanse:
@adgad


TheoLeanse:
@i-like-robots any point in this or shall I close

i-like-robots:
can't be that important!

PR #942, opened 2017-05-19
Ads - updating metrics for new beacon query

andrewgeorgiou1981:
@andygnewman @fenglish


andygnewman:
Looks like the right event to trigger it from.
Just a couple of linting things to check before merging in.

PR #944, opened 2017-05-23
add barrierReferrer query param to barrier view tracking context

quarterto:
@tavvy


quarterto:
ah i didn't have write access so i forked it, but CI doesn't run on forks. could you give me write access and i'll redo the PR @tavvy

PR #951, opened 2017-05-26
use capi2 id for ads when available

wheresrhys:
@andygnewman @andrewgeorgiou1981


wheresrhys:
image

andygnewman:
Great, this was going to be my next step, thanks.

Looks like a test needs fixing though.

PR #957, opened 2017-05-30
Check webpack build on test

leggsimon:
A couple of releases happened that failed once merged because it only failed during the `webpack --bail --config build/deploy/webpack.config.js` command which lived within `make deploy`.
`make deploy` only happened on master and therefore it didn't fail on branches. This should ensure we catch these errors on branch builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,9 @@ copy-stylesheet-partial:
 build-css-loader:
 	uglifyjs browser/layout/src/css-loader.js -o browser/layout/partials/css-loader.html
 
+build-webpack:
+	webpack --bail --config build/deploy/webpack.config.js
+
 test-server-coverage: ## test-server-coverage: Run the unit tests with code coverage enabled.
 	istanbul cover node_modules/.bin/_mocha --report=$(if $(CIRCLECI),lcovonly,lcov) server/test/*.test.js server/test/**/*.test.js
 
@@ -77,7 +80,7 @@ a11y: test-build pally-conf
 
 # Note: `run` executes `node demo/app`, which fires up express, then deploys
 # a test static site to s3, then exits, freeing the process to execute `nightwatch a11y`.
-test: developer-note verify pally-conf test-server test-browser test-build test-webpack run nightwatch a11y
+test: developer-note verify pally-conf test-server test-browser test-build test-webpack run nightwatch a11y build-webpack
 
 developer-note:
 ifeq ($(NODE_ENV),) # Not production
@@ -92,7 +95,7 @@ endif
 test-dev: verify test-browser-dev test-webpack
 
 deploy:
-	webpack --bail --config build/deploy/webpack.config.js
+	make build-webpack
 	node ./build/deploy/s3.js
 	$(MAKE) build-css-loader
 	$(MAKE) npm-publish
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,9 @@ endif
 build:
 	webpack --config demo/webpack.config.js --dev
 
+build-production:
+	build-bundle
+
 watch:
 	webpack --config demo/webpack.config.js --dev --watch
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -92,7 +95,7 @@ endif
 test-dev: verify test-browser-dev test-webpack
 
 deploy:
-	webpack --bail --config build/deploy/webpack.config.js
+	make build-webpack
```
adambraimbridge: Should this be `$(MAKE) build-webpack`?

bjfletcher: `deploy: build-webpack` is more Make-style

Makefile
```diff
@@ -77,7 +80,7 @@ a11y: test-build pally-conf
 
 # Note: `run` executes `node demo/app`, which fires up express, then deploys
 # a test static site to s3, then exits, freeing the process to execute `nightwatch a11y`.
-test: developer-note verify pally-conf test-server test-browser test-build test-webpack run nightwatch a11y
+test: developer-note verify pally-conf test-server test-browser test-build test-webpack run nightwatch a11y build-webpack
```
bjfletcher: The CircleCI app preset says to run during test: `make build-production test provision`

I wonder if it'd make better sense to do this instead:

```
build-production: build-webpack
```

and remove the `webpack --bail --config build/deploy/webpack.config.js` build line in the `deploy` target.

So it's in keeping with how we do it elsewhere in Next.

Disclaimer: I've barely worked with `n-ui` and if `n-ui` needs to be different then cool 👍

Makefile
leggsimon: @bjfletcher will omitting `make` or `$(MAKE)` work with other non-make commands. The whole `deploy` command looks like 
```
 deploy:
 	make build-webpack
  	node ./build/deploy/s3.js
  	$(MAKE) build-css-loader
  	$(MAKE) npm-publish
```

Makefile
leggsimon: Actually, I think I'd rather it just be part of the deploy step. I'd be nervous about doing the `build-production` and then `test` incase `test` generates stuff we don't want to be deployed.

Makefile
leggsimon: So TIL, `$(MAKE)` creates a new instance of make and then runs that command. we don't need to do that

Makefile
bjfletcher: @leggsimon `test` task should absolutely not generate anything for deployment... it should only test
I think all developers are clear about that
Anything generated for deployment should be in the `build` step, never `test`.
If you see any project that generates/builds in the `test` step then that project should really move it into `build`.

leggsimon: Have updated

Makefile
```diff
@@ -61,6 +64,17 @@ copy-stylesheet-partial:
...
+build-bundle:
+	webpack --bail --config build/deploy/webpack.config.js
+
+deploy-s3:
+	node ./build/deploy/s3.js
+
+rebuild-user-facing-apps:
+	# only autodeploy all apps in office hours
+	HOUR=$$(date +%H); DAY=$$(date +%u); if [ $$HOUR -ge 8 ] && [ $$HOUR -lt 16 ] && [ $$DAY -ge 0 ] && [ $$DAY -lt 6 ]; then \
+	echo "REBUILDING ALL APPS" && sleep 20 && nht rebuild --all --serves user-page; fi
```
bjfletcher: "In 20 seconds we will start rebuilding ALL apps... quit now if you don't want to!"? :)

components/n-ui/tracking/third-party/chartbeat.js
```diff
@@ -6,7 +6,7 @@ Eventually this will be replaced by FT's in-house "lantern" application, utilisi
 */
 
 import {cookieStore} from 'n-ui-foundations';
-import {loadScript} from '../../../../browser/bootstrap/js/utils';
+import {loadScript} from '../../../../browser/js/utils';
```
bjfletcher: a separate bug?

leggsimon: Yep, this is actually already in master but for some reason has been added in this PR. A rebase gone awry I think.

Makefile
leggsimon: https://github.com/Financial-Times/n-ui/issues/972

PR #962, opened 2017-05-31
WIP - Webpack 2

leggsimon:
## Tasks:

- [x] Move spread out config into one larger config file
- [x] Change all current usage to new config
- [x] Bump the Webpack version
- [x] move the ftNextUI stuff into wrapper.js
- [x] Review plugins delete/bump where necessary
- [x] Comment what each part of the config is actually doing

## For later
- [ ] Try and standardise apps' entry points
- [ ] Improve how `entry` properties are configured
- [ ] Review whether React plugins are still necessary
- [ ] Look into building an ES6 bundle


build/webpack/webpack.config.js
```diff
@@ -0,0 +1,170 @@
...
+
+
+const handlebarsConfig = () => {
+	const extraHelperDirs = glob.sync('**/node_modules/@financial-times/**/handlebars-helpers')
+		.map(dir => path.resolve(dir));
+	return {
+		debug: false, // set to true to debug finding partial/helper issues
+		extensions: ['.html'],
+		helperDirs: [
+			path.resolve('./node_modules/@financial-times/n-handlebars/src/helpers'),
```
JakeChampion: Why not use `require.resolve` instead of `path.resolve`?  https://nodejs.org/dist/latest-v6.x/docs/api/globals.html#globals_require_resolve

build/webpack/webpack.config.js
```diff
@@ -0,0 +1,170 @@
...
+
+const path =require('path');
+const glob = require('glob');
+const BowerWebpackPlugin = require('bower-webpack-plugin');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const autoprefixer = require('autoprefixer');
+
+
+const handlebarsConfig = () => {
+	const extraHelperDirs = glob.sync('**/node_modules/@financial-times/**/handlebars-helpers')
```
JakeChampion: This looks like some magic. You are finding all handlebars-helpers directories in all ft npm modules and adding them to your handlebars configuration. I'm assuming that this might cause issues if two handlebars helpers had the same function name, the last defined function overrides all the previous ones, right?

build/webpack/webpack.config.js
```diff
@@ -0,0 +1,170 @@
...
+
+module.exports = () => ({
+	devtool: 'source-map',
+
+	resolve: {
+		root: [
+			path.resolve('./bower_components'),
+			path.resolve('./node_modules')
+		],
+		alias: Object.assign(require('babel-polyfill-silencer/aliases'), {
```
JakeChampion: `babel-polyfill-silencer/aliases` is not required if using the newest version of Babel by the way 🎉

build/webpack/webpack.config.js
```diff
@@ -0,0 +1,170 @@
...
+							{ polyfill: false }
+						],
+						// require.resolve('babel-plugin-transform-es2015-modules-commonjs'), // not sure this was ever actually used??
+						[
+							require.resolve('babel-plugin-transform-es2015-classes'),
+							{ loose: true }
+						]
+					],
+					presets: [
+						require.resolve('babel-preset-react'),
```
JakeChampion: Are Next still using JSX syntax? This also pulls in other features such as object spread etc.

build/webpack/webpack.config.js
```diff
@@ -0,0 +1,170 @@
...
+					/node_modules/
+				],
+				query: {
+					babelrc: false, // ignore any .babelrc in project & dependencies
+					cacheDirectory: true,
+					plugins: [
+						require.resolve('babel-plugin-add-module-exports', true),
+						[
+							require.resolve('babel-plugin-transform-runtime'),
+							{ polyfill: false }
```
JakeChampion: If you change this to:
```
{
	helpers: false,
	polyfill: false,
	regenerator: true,
}
```
Then you should not get any core-js polyfills in your code 🎉

build/webpack/webpack.config.js
```diff
@@ -0,0 +1,170 @@
...
+			flexbox: 'no-2009'
+		}) ];
+	},
+
+	plugins: [
+		new BowerWebpackPlugin({
+			includes: /\.js$/,
+			modulesDirectories: path.resolve('./bower_components')
+		}),
+		new ExtractTextPlugin('[name]'),
```
JakeChampion: Not quite sure what this is doing

test/webpack.spec.js
```diff
@@ -1,7 +1,9 @@
 const expect = require('chai').expect;
-const nUiWebpack = require('../build/webpack');
+const nUiWebpack = require('../build/webpack/webpack.config.js');
 
-describe('nUiWebpack', () => {
+// TODO: make this test good, if we can. maybe we can't? ¯\_(ツ)_/¯
```
JakeChampion: What are you hoping to test?

build/webpack/webpack.config.js
leggsimon: 👍 Yep, I'll look at improving this later

build/webpack/webpack.config.js
leggsimon: Not as far as I know, this could possibly go now

leggsimon: Oh actually, KAT uses React and I think the Gift Article thing might still be React

build/webpack/webpack.config.js
leggsimon: 🤷‍♂️

test/webpack.spec.js
leggsimon: This was brought over from `n-webpack` it possibly won't be as necessary with one large config. I guess it might've been more useful when there was a bit more logic in building the config?

JakeChampion:
If I could approve this PR, I would 👍

PR #963, opened 2017-06-01
removed Lionel face from the slider

jkerr321:
Now it look like dis
image


leggsimon:
Could you remove this as well then please 😊 
https://github.com/Financial-Times/n-ui/blob/master/components/n-ui/subscription-offer-prompt/main.scss#L89-L107

PR #966, opened 2017-06-02
Refactor app build

leggsimon:


build/app/webpack.js
```diff
@@ -43,33 +39,48 @@ const baseConfig = require(path.join(process.cwd(), 'n-ui-build.config.js'));
...
+const webpackConfigs = [];
+
+/*
+We no longer build a main.js for the app when generating the standard asset variants
+so this config is for all entry points defined by an app *excluding* the main.js one
+
+Mostly this config will only be for main.css.
+*/
+const nonMainJsWebpackConfig = nWebpack();
+nonMainJsWebpackConfig.entry = filterEntryKeys(baseConfig.entry, /main\.js$/, true);
```
wheresrhys: you could just `delete entry['main.js']` if we're thinking of moving down the mroe standardised approach

PR #969, opened 2017-06-08
Change ads lazyload viewport margin

fenglish:
See this ticket below;
https://trello.com/c/PSwSNRFX/143-optimise-ad-lazy-load-trigger-on-mobile

### DIFF
--- a/components/n-ui/ads/test/oAdsConfig.spec.js
+++ b/components/n-ui/ads/test/oAdsConfig.spec.js
@@ -27,6 +27,7 @@ describe('Config', () => {
 	});
 
 	describe('third party', () => {
+
 		it('Should set gpt configuration value according to app name', () => {
 			const flags = { get: () => true };
 			const config = oAdsConfig(flags, 'article', false);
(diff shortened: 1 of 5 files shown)


components/n-ui/ads/test/oAdsConfig.spec.js
```diff
@@ -103,8 +105,111 @@ describe('Config', () => {
...
+
+			beforeEach(() => {
+				global.innerWidth = 759;
+			});
+
+			afterEach(() => {
+				delete global.innerWidth;
+			});
+
+			it('Should pass 50% when the flag\'s value is 50', () => {
```
wheresrhys: as a time saving approach to future tests you can do something like the following
```
['50', '100', '150']
  .forEach(margin => {
    it(`Should pass ${margin}% when the flag\'s value is ${margin}`, () => {
        // write the test, using the margin value as appropriate
    })
  })
```
No need to do it for this PR (unless you want to), but thought I'd point it out as it's easy to think of tests as being somehow restricted in the way they're written, but actually you can use any js techniques you like to avoid code repetition.

PR #971, opened 2017-06-12
Bump webpack version

leggsimon:
## ❗️ Please see [this comment](https://github.com/Financial-Times/n-ui/pull/971#issuecomment-309573160) below

Part of this PR removes the wrapping functionality from webpack. I have just commented it out for the time being so we can have a proper discussion in the future about where this should live. In my opinion it adds an unnecessary amount of "magic" and a few other devs in next agreed and didn't like the idea that it was adding code outside of the app itself. @wheresrhys has different opinions and thinks it should live in n-ui.

My proposal is that instead of having webpack wrap an app's JS in a kind of `init` function we should follow the approach that Origami take with all of their components [(example here)](https://github.com/Financial-Times/o-forms/blob/master/main.js) and have each app be responsible for that.

**Example of the breaking change needed:**
```js
// main.js

function initApp () {

	// the current content of an app's `main.js`

	document.removeEventListener('ftNextnUiLoaded', initApp);
}

if (window.ftNextnUiLoaded) {
	initApp();
} else {
	document.addEventListener('ftNextnUiLoaded', initApp);
}
```


build/app/webpack.js
```diff
@@ -49,6 +49,7 @@ Mostly this config will only be for main.css.
 */
 const nonMainJsWebpackConfig = nWebpack();
 nonMainJsWebpackConfig.entry = filterEntryKeys(baseConfig.entry, /main\.js$/, true);
+console.log('\x1b[36m', nonMainJsWebpackConfig.entry, '\x1b[0m')
```
JakeChampion: Rogue `console.log`

build/deploy/webpack.config.js
```diff
@@ -33,7 +33,7 @@ module.exports = configs
 			const webpack = require('webpack');
 			// TODO: figure out what these are for and describe them
 			webpackConfig.plugins.push(new webpack.DefinePlugin({ 'process.env': { 'NODE_ENV': '"production"' } })) /* Only if production */
-			webpackConfig.plugins.push(new webpack.optimize.UglifyJsPlugin({ 'compress': { 'warnings': false } }))
+			webpackConfig.plugins.push(new webpack.optimize.UglifyJsPlugin({}))
```
JakeChampion: Why are we passing an empty object here?

Makefile
```diff
@@ -92,7 +92,7 @@ endif
 test-dev: verify test-browser-dev test-webpack
 
 deploy:
-	webpack --bail --config build/deploy/webpack.config.js
+	webpack --bail --config build/deploy/webpack.deploy.config.js --define process.env.NODE_ENV="'production'"
```
wheresrhys: is there a reason we can't just `export NODE_ENV=production` on the previous line

build/webpack/webpack.config.js
```diff
@@ -5,9 +5,9 @@ As well as just making it more accessible without having to dive around all the
 transforms. Would prefer to have a well commented file like Origami:
 */
 
-const path =require('path');
+const path = require('path');
 const glob = require('glob');
-const BowerWebpackPlugin = require('bower-webpack-plugin');
+// const BowerWebpackPlugin = require('bower-webpack-plugin');
```
wheresrhys: maybe rename this file to reflect which of the build types it's for

Makefile
leggsimon: Yes, so this setting invokes the webpack `DefinePlugin`. What that does is goes through the code and replaces any instance of `process.env.NODE_ENV` with the string `'production'` so that 
```js
if (process.env.NODE_ENV === 'production') { /* stuff */ }
```
becomes
```js
if ('production' === 'production') { /* stuff */ }
```
which becomes
```js
if (true) { /* stuff */ }
```
and then the uglify step just removes the `if` statement completely to become just 
```
/* stuff */
```

build/app/cli.js
```diff
@@ -50,7 +50,7 @@ program
 
 		devAdvice();
 
-		shellpipe(`webpack ${options.production ? '--bail' : '--dev'} --config ${webpackConfPath}`)
+		shellpipe(`webpack --bail --config ${webpackConfPath}`)
```
leggsimon: Ahh, I _do_ need to deal with this.

leggsimon:
I'll keep the above description in for reference but currently I have gone back and decided against removing the `Wrap` code and introducing a breaking change, just yet. The main reason for this is that AFAIK this won't play nicely with apps that use AMD modules over CommonJS in the client side JS. We don't _have_ to introduce the breaking change right now and would rather this PR didn't actually change any current behaviour.

build/app/webpack.app.common.js
```diff
@@ -0,0 +1,152 @@
...
+		descriptionFiles: ['bower.json', 'package.json'],
+
+		// These fields in the description files are looked up when trying to resolve the package directory
+		mainFields: ['main', 'browser'],
+
+		// These files are tried when trying to resolve a directory
+		mainFiles: [
+			'index',
+			'main',
+			'src/main-client' /* HACK: this is becaause of n-image's entrypoin not being main.js or index.js */
```
JakeChampion: 🙈

leggsimon: 😏

PR #973, opened 2017-06-12
Ads optimize lazyload viewportMargin tablet

fenglish:
See this ticket below;
https://trello.com/c/zGJOxLIq/142-optimise-ad-lazy-load-trigger-on-tablet

and I have deleted a test `Should pass 0% when the flag\'s value is control` from mobile(Small size) as well because `flags.get('adOptimizeLazyLoadSmall')` just returns false when the variant is control.


andrewgeorgiou1981:
👍  looks good

PR #978, opened 2017-06-16
removed n-marketing

jkerr321:
removed `n-marketing` from `n-ui` and have directly initialised it in `front-page` and `stream-page` instead


wheresrhys:
woo with some yay on the side :tada:

PR #990, opened 2017-06-26
add dist/main to module entrypoints

leggsimon:
we need to think of a better way to resolve these in the future but for the time being this should allow stream-page to not lag behind the current n-ui


leggsimon:
https://github.com/Financial-Times/n-ui/issues/991

PR #998, opened 2017-06-30
Rhys/multiple bundles

wheresrhys:
So this is an optimistic illustration of what I think it'd be really good to do

- o-errors is huge and rarely changes, so worth making into its own separately cached bundle
- font-loader is tiny and, as a separate bundle, can be cached very well, and it's really important to load it quickly, so also worth splitting out
 
Raises lots of questions about replicating this prod bundle behaviour in the demo, when bower linking into an app and all the other cases. I wonder if it'd be possible to somehow get rid of some of the differences between the different builds

Should we be using webpack's code-splitting for this?

TODO:
- [x] avoid font-loader importing all of n-ui-foundations
- [x] tests for additional files
- [ ] tests for defer but not async on scripts
- [ ] should o-errors and font-loader be loaded async?
- [x] move everything simple backout of webpack-entry-points.js
- [x] comment how n-ui imported in the demo app

Measure success against this https://beacon.ft.com/data/query-wizard?query=page%3Aload-timing-%3Emedian(context.timings.marks.fontsLoaded)-%3Egroup(device.primaryHardwareType)-%3Efilter(device.primaryHardwareType%3FDesktop%2CMobile%20Phone)-%3Efilter(device.primaryHardwareType)-%3ErelTime(previous_70_weeks)-%3Einterval(w)-%3Edivide(1000)


wheresrhys:
I'm now pretty optimistic we can have a tiny number of webpack configs

- a utility detects if the 'app shell' is to be loaded from ./browser/bundles or ./bower_components/n-ui/browser/bundles. 
- demo build can then basically become an app build with slightly different entry points, and with `NEXT_APP_SHELL=local` always on
- As a consequence of this, the demo build, with production mode on, contains the deploy build as a subset of itself

Still a bit of ironing out to do around treatment of css and file names, but doubt it's insurmountable @leggsimon

PR #1002, opened 2017-07-03
use deferred scripts instead of custom async script handling

wheresrhys:
Since we no longer support ie10 for js, we don't need to worry about not supporting defer http://caniuse.com/script-defer

As defer gives us the loading behaviour we need ([async load, but execute in predictable order](https://peter.sh/experiments/asynchronous-and-deferred-javascript-execution-explained/)) we can get rid of some (maybe eventually all) our loading hacks


build/app/webpack.js
```diff
@@ -71,14 +70,7 @@ const nUiExternal = require('../../browser/js/webpack-entry');
 const nUiExternalPoints = nUiExternal(baseConfig.nUiExcludes);
 const mainJsWebpackConfig = webpackMerge(commonAppConfig, {
 	entry: modifyEntryKeys(baseConfig.entry, /main\.js$/, name => name.replace(/\.js$/,'-without-n-ui.js')),
-	externals: nUiExternalPoints,
-	plugins:[
-		new Wrap(
```
JakeChampion: 🙌

PR #1003, opened 2017-07-03
improved ignore rules

wheresrhys:


.bowerignore
```diff
@@ -0,0 +1,22 @@
+/.idea/
```
i-like-robots: should this be covered by people's global gitignore?

secret-squirrel.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
 	files: {
-		allow: [],
+		allow: ['.bowerignore'],
```
i-like-robots: npmignore too?

wheresrhys: already allowed in the global rules 🤷‍♂️

.bowerignore
wheresrhys: yep

PR #1015, opened 2017-07-06
removed n-marketing from bower.json

jkerr321:
n-marketing doesn't need to be listed as a bower dependency in `n-ui` as it's installed directly in `next-front-page` and `next-stream-page`


wheresrhys:
indeedy

PR #1018, opened 2017-07-07
fix(redirect): do not redirect to barrier pages

doramatadora:
Once user signs in successfully from a barrier page, redirect to `/` rather than back to the barrier
 🐿2.5.10


server/models/navigation/navigationModelV2.js
```diff
@@ -66,7 +66,11 @@ module.exports = class NavigationModelV2 {
 		const currentPathName = url.parse(currentUrl).pathname;
 		for(let item of navData.items){
 			if(typeof item.url === 'string' && item.url.includes('${currentPath}')){
-				item.url = item.url.replace('${currentPath}', currentUrl);
+				if(!currentPathName || currentPathName.indexOf('/products') === -1 || currentPathName.indexOf('/barriers') === -1) {
```
remybach: A regex `.test` here would be more concise I think.

PR #1020, opened 2017-07-07
Set default font to metric to fix font issues

GlynnPhillips:


wheresrhys:
Baddabing badda...
image

PR #1029, opened 2017-07-12
Add ids to subnav menu items

remybach:
I need to be able to point a tooltip to a particular menu item.


remybach:
It can't be a data attribute unfortunately as `o-tooltip` gets the target element by id (since it can only have a singular target and giving it an id enforces this).

I figured I could get away with not having it behind some kind of flag since the behaviour is opt in (i.e.: you have to pass in an id for the particular item you need it for), and creating a whole new flag seemed overkill.

I'll add a comment briefly explaining why it's there though 👍

remybach:
There is some work going into `o-tooltip` that means I probably won't be needing this after all.

PR #1035, opened 2017-07-17
allow different product name

lc512k:
@wheresrhys @leggsimon 
KAT and Syndication use n-ui but need a way of changing the product name, hardcoded until now.


leggsimon:
Is KAT and Syndication not part of next?

lc512k:
@leggsimon They are part of the Next team, but they are distinct products from FT.com. Also, for analytics we need it to have a different product code. One example of why is wanting to count all the page views on Next, as in FT.com. We wouldn't want KAT and Syndication page views in that number.

PR #1037, opened 2017-07-18
Add iOS smart banner and add subscriberCohort logic

remybach:
🐿2.5.10


browser/layout/wrapper.html
```diff
@@ -33,10 +33,24 @@
...
+		{{#unless @root.flags.disableAndroidSmartBanner}}
+		<link rel="manifest" href="/assets/manifest/manifest-v5-app-banner.json">
+		{{/unless}}
+		{{/if}}
+
 		{{#if @root.flags.disableAndroidSmartBanner}}
 		<link rel="manifest" href="/assets/manifest/manifest-v5.json">
-		{{else}}
-		<link rel="manifest" href="/assets/manifest/manifest-v5-app-banner.json">
+		<script>
```
tavvy: If this disables the banner, then could we not always load `/assets/manifest/manifest-v5-app-banner.json` and just put this script behind the flag?

remybach: That's a good point actually... will do some investigation into how this will affect stuff like manually choosing to "add to homescreen"

PR #1044, opened 2017-07-25
Specify a product value other than 'next' for spoor

dapenguin:
Changes to allow us to specify a different `product` value to send to spoor. Defaults to sending 'next' as the value.


lc512k:
FYI @leggsimon @wheresrhys backwards compatible with our current way of working in next and won't add any extra markup, except for KAT and anything else that specifies a product. But please shout if something's off.

wheresrhys:
where does this get passed to spoor?

dapenguin:
@wheresrhys It's passed in via the `context`, which is set here: [master/components/n-ui/tracking/ft/index.js#L44](https://github.com/Financial-Times/n-ui/blob/master/components/n-ui/tracking/ft/index.js#L44)

PR #1055, opened 2017-07-31
remove adInitEarlierNui

fenglish:
We didn't get a big difference with viewabilty when ads are initialised earlier. We are going to take the flag off and go back before we added the code.
 
🐿 v2.5.13


fenglish:
I'm sorry this PR is other way around:sob:. I should keep the feature which ads are initialised ealier even we didn't get a big impact. I'm going to make new PR.

PR #1056, opened 2017-07-31
remove adInitEalierNui flag

fenglish:
We are going to roll out the feature which ads are initialised earlier. This PR is the correct one!

This PR is the opposite way of the PR below. (I misunderstood the situation and made the PR)
https://github.com/Financial-Times/n-ui/pull/1055
 
🐿 v2.5.13


andrewgeorgiou1981:
👍

PR #1062, opened 2017-08-09
feat(comms): hide b2c comms for b2b

doramatadora:
* hide the header promo from B2B users
* bump node
* bump some deps (I think `proptypes` specifically might reduce some of the Sentry / front page errors)


doramatadora:
@leggsimon only affecting breaking changes in `sinon`, `chai` and `karma-sauce-launcher`, so I left those as they were. GTG?

PR #1067, opened 2017-08-18
Add the desktop app banner

remybach:


tavvy:
looks good once the css file size is addressed in the component as discussed.

remybach:
For posterity, @leggsimon's point about the added 4kb is due to `o-typography` not being included silently. I'll investigate preventing this happening after this has gone in.

PR #1068, opened 2017-08-18
Excludes underscore in node modules from using babel-loader

dapenguin:
Excludes underscore in node modules from using babel-loader due to the following issue: https://stackoverflow.com/questions/39301399/underscore-gives-error-when-bundling-with-webpack

 🐿 v2.5.16


laurieboyes:
We've now moved away from the dependency that relied on underscore, so this PR is no longer necessary

PR #1073, opened 2017-08-23
add es2016 and 2017 babel presets

wheresrhys:
🐿2.5.10


hzoo:
fyi I would swap out es2015,16,17 for https://github.com/babel/babel-preset-env (it's equivalent).

wheresrhys:
cool, thanks for the tip.

PR #1074, opened 2017-08-23
Remove ternary and attach headline variant to page only

missamynicholson:
@tom-parker

### DIFF
--- a/components/n-ui/tracking/ft/index.js
+++ b/components/n-ui/tracking/ft/index.js
@@ -106,10 +106,16 @@ const oTrackingWrapper = {
 				useSendBeacon: flags.get('sendBeacon')
 			});
 
-			//headline testing
-			const headlineTesting = document.querySelector('[data-trackable-headline-variant]');
-			if (headlineTesting) {
-				pageViewConf.context['headline-variant'] = headlineTesting.getAttribute('data-trackable-headline-variant');
+			//headline testing, add variant to the page view event as long as there is only one article under test
+			if (location.pathname === '/') {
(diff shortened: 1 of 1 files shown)


tom-parker:
👍 

Unfortunately I think the logic in o-tracking that moves 'headline-test' and the relevant value into the context is going to change so this is likely to break in the future

components/n-ui/tracking/ft/index.js
```diff
@@ -104,9 +106,17 @@ const oTrackingWrapper = {
...
-			//headline testing
-			const headlineTestingVariant = document.querySelector('[data-trackable-headline-variant]') ? document.querySelector('[data-trackable-headline-variant]').getAttribute('data-trackable-headline-variant') : null;
-			context['headline-variant'] = headlineTestingVariant;
+			//headline testing, add variant to the page view event as long as there is only one article under test
+			if (location.pathname === '/') {
+				const alternativeHeadlines = document.querySelectorAll('[data-trackable-headline-variant]');
+				let hrefs = [];
+				alternativeHeadlines.forEach(headline => hrefs.push(headline.getAttribute('href')));
+
+				const isOnlyOneArticle = hrefs.length ? !!hrefs.reduce((a, b) => a === b ? a : NaN) : false;
```
tom-parker: Maybe I've missed something but why `NaN` here?

missamynicholson: Good point, could have just been false.
Have updated it to slice the NodeList as you suggested, and changed the href equality check to use Array.every() as it's a bit neater.

tom-parker:
👍

PR #1083, opened 2017-09-06
update host because CORS

gvonkoss:
Fixes https://jira.ft.com/browse/NFT-910 which was pointing at next-errors.ft.com/search... which wasn't anything. 

cc/ @JakeChampion @i-like-robots

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
   "bundlesize": [
     {
       "path": "./public/n-ui/es5.js",
-      "threshold": "56 Kb"
+      "threshold": "56.5 Kb"
     },
     {
       "path": "./public/n-ui/font-loader.js",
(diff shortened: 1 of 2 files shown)


wheresrhys:
I'll be merging https://github.com/Financial-Times/n-ui/pull/1043 soon. Could you port this fix to https://github.com/Financial-Times/n-topic-search too please

I also wonder if a better fix would be to rethink error pages a bit, but that's just one to chalk up for later

JakeChampion:
@gvonkoss It looks like Circle wants the file to be smaller :/
>FAIL  ./public/n-ui/es5.js: 56.13kB > maxSize 56kB gzip

gvonkoss:
yeah but I don't know how to change that. And I don't know if the n-express warning has anything to do with it

package.json
```diff
@@ -121,7 +121,7 @@
   "bundlesize": [
     {
       "path": "./public/n-ui/es5.js",
-      "threshold": "56 Kb"
+      "threshold": "56.5 Kb"
```
JakeChampion: 😂

gvonkoss: 😬

PR #1086, opened 2017-09-13
make n-ui work again when flags disabled

doramatadora:
```js
app = express({ ... withFlags: false ... }) 
```
throws right now because `withFlags: false` means no `res.locals.flags`

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -67,9 +67,9 @@ module.exports = options => {
 
 	// set whether or not to disable the app install banner.
 	app.use(function (req, res, next) {
-		app.locals.__disableAndroidBanner = (!res.locals.flags.subscriberCohort || res.locals.flags.disableAndroidSmartBanner);
-		app.locals.__disableIosSmartBanner = (!res.locals.flags.subscriberCohort || res.locals.flags.disableIosSmartBanner);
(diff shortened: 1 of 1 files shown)


wheresrhys:
Why d'you need n-ui to work when flags is disabled?

doramatadora:
@rhys because I want to use it in a non-next app, and the option used to work before those changes.

doramatadora:
ok @wheresrhys but it should still work when there are no flags, on principle

wheresrhys:
not really - all next front ends should have flags enabled, otherwise standard features that are behind a flag (e.g. search suggestions, tracking) won't work

doramatadora:
@wheresrhys If you aren't going to safeguard against `withFlags: false` (or `withFlags` being an option at all), then you should safeguard for the TypeError that results out of getting properties of an undefined `res.locals.flags`, which in turns makes everything **not run**. This is common practice throughout the code: https://github.com/Financial-Times/n-ui/blob/26cba144e7a78a851d84cfd3729e1881c938f67b/server/models/anon.js#L23 so am unsure – are you objecting to my PR or something else?

wheresrhys:
that code is probably a hangover from when the code was originally imported from n-express, which is supposed to support turning off flags.

Perhaps n-ui should force flags to be on, which removes the ability to get into cryptic errors like you describe, but that would technically be a breaking change, so one for v7.

PR #1088, opened 2017-09-14
Remove role presentation from the header

GlynnPhillips:
This will soon cause errors for everyone when Pa11y runs. The
`role="presentation"` attribute was only ever really added to make sure
we were hiding it from keyboard focus and from the accessibility api.
However we also use `aria-hidden="true"` on the parent and
`tabindex="-1"` on all the children to achieve this.

 🐿2.5.10


i-like-robots:
🆗 🚢

PR #1094, opened 2017-09-27
Ads - specifying ad loading viewport margins for different screensizes/page types

andrewgeorgiou1981:
@adgad @fenglish @alexflorisca

### DIFF
--- a/components/n-ui/ads/js/oAdsConfig.js
+++ b/components/n-ui/ads/js/oAdsConfig.js
@@ -56,6 +56,11 @@ module.exports = function (flags, appName, adOptions) {
 	}
 
 	function setViewportMarginBy (variant) {
+
+		if (appName.toLowerCase().substr(0, 3) === fro && res=small) return '15%';
+		// (pagetype === fro && res=med) return '5%';
+		// (pagetype === str && res=small) return '5%';
+		// (pagetype) === str && res=med) return '15%';
 		switch (variant) {
 			case '5':
 				return'5%';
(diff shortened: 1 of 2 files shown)


components/n-ui/ads/js/oAdsConfig.js
```diff
@@ -55,20 +55,21 @@ module.exports = function (flags, appName, adOptions) {
...
-				return '10%';
-				break;
-			case '15':
-				return '15%';
-				break;
-			default:
-				return '0%';
+	function setViewportMarginBy () {
+		let pt = appName;
+		pt = pt.toLowerCase().substr(0, 3);
```
adgad: Is it necessary to substring it here? Might be more readable later on to refer to 'front-page' (or whatever the full name is)

components/n-ui/ads/js/oAdsConfig.js
```diff
@@ -55,20 +55,21 @@ module.exports = function (flags, appName, adOptions) {
...
+		let scrnSize = utils.getScreenSize();
+		if (pt === 'fro') {
+			if (scrnSize < 760) {return '15%';}
+			else {
+				if (pt === 'fro' && scrnSize < 980) {return '5%';}
```
adgad: checking 'fro' again here shouldn't be necessary

components/n-ui/ads/test/oAdsConfig.spec.js
```diff
@@ -141,36 +141,80 @@ describe('Config', () => {
...
-
-			beforeEach(() => {
-				sandbox.stub(utils, 'getScreenSize', () => { return 979; });
-			});
-
-			afterEach(() => {
-				sandbox.restore();
-			});
+		it('Should pass 15% when screen size is narrower than 760px and adOptimizeLazyLoad flag is defined and appName is fro', () => {
+			sandbox.stub(utils, 'getScreenSize', () => { return 750; });
```
adgad: We should probably remove the flag from the tests...it's a little confusing otherwise

andrewgeorgiou1981: i was planning to keep the flag but update it so that its no longer an MVT - this would allow us to toggle the viewport margin. However, i guess it doesnt give us much benefit so, as you say it might be less confusing to remove it.

PR #1095, opened 2017-09-27
add nui build-sass

doramatadora:
Just an idea: could we have a `nui build-sass` command? It would be great to have a way to build css only locally. Doing it concurrently with webpack hangs for a lot of us.

Or a `--sass-only` flag on `nui build`?

### DIFF
--- a/build/app/cli.js
+++ b/build/app/cli.js
@@ -74,8 +74,8 @@ program
 			}
 		}
 
-		concurrentCommands = concurrentCommands.length 
-			? concurrentCommands 
+		concurrentCommands = concurrentCommands.length
+			? concurrentCommands
 			: Object.keys(commands).map(c => command[c]);
 
 		shellpipe(`concurrently ${concurrentCommands.join(' ')}`)
@@ -87,7 +87,7 @@ program
 				const buildTime = Date.now() - buildStartTime;
 
 				// Don't send metrics from CircleCI builds
-				if (!process.env.CIRCLECI) {
+				if (!process.env.CIRCLECI && !options['js-only'] && !options['sass-only']) {
 					sendBuildMetrics(appPackageJson.name, buildTime);
 				}
 
--- a/build/app/cli.js
+++ b/build/app/cli.js
@@ -54,49 +54,29 @@ program
 	.command('build')
 	.description('Builds n-ui apps, ready to be deployed to your favourite s3 bucket or heroku host')
 	.option('--production', 'Builds with production settings')
+	.option('--sass-only', 'Builds Sass only')
+	.option('--js-only', 'Builds JavaScript only')
 	.action(options => {
 
 		devAdvice();
 		const buildStartTime = Date.now();
-		const cssBuildCommands = cssEntryPoints.map(([target, entry]) => {
-			const script = './node_modules/@financial-times/n-ui/scripts/build-sass.sh';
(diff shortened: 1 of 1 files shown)


leggsimon:
Are you proposing the same for JS as well?

Just in case you weren't aware, `nui watch` will rebuild sass only on save if you are editing that.

alexnaish:
The issue for me isn't whats built during the `watch` task, its the problem that 9/10 times I just need to build `head.css` in order to run my unit-tests for example and both `make build` and `make watch` often fail to build SASS first time.

leggsimon:
I think I would prefer a `--sass-only`/`--js-only` option to be honest @doramatadora

build/app/cli.js
```diff
@@ -54,16 +54,29 @@ program
...
+	.option('--js-only', 'Builds JavaScript only')
 	.action(options => {
 
 		devAdvice();
 		const buildStartTime = Date.now();
-		const cssBuildCommands = cssEntryPoints.map(([target, entry]) => {
-			const script = './node_modules/@financial-times/n-ui/scripts/build-sass.sh';
-			return `'${script} ${entry} ${target}'`;
-		}).join(' ');
+		let concurrentComands = [];
```
alexnaish: Picky (sorry!) but commands has 2 `m`s

build/app/cli.js
```diff
@@ -54,49 +54,29 @@ program
...
-				const buildTime = Date.now() - buildStartTime;
+		const script = './node_modules/@financial-times/n-ui/scripts/build-sass.sh';
+		const commands = {
+			'js-only': `'webpack --bail --config ${webpackConfPath} ${options.production ? '-p' : ''}'`,
+			'sass-only': cssEntryPoints.map(([target, entry]) => `'${script} ${entry} ${target}'`).join(' ')
+		};
 
-				// Don't send metrics from CircleCI builds
-				if (!process.env.CIRCLECI) {
-					sendBuildMetrics(appPackageJson.name, buildTime);
```
leggsimon: Can we keep this in please for a full build.

build/app/cli.js
```diff
@@ -54,49 +54,29 @@ program
...
-	.option('--production', 'Builds with production settings')
-	.action(options => {
-
-		devAdvice();
-		const buildStartTime = Date.now();
-		const cssBuildCommands = cssEntryPoints.map(([target, entry]) => {
-			const script = './node_modules/@financial-times/n-ui/scripts/build-sass.sh';
-			return `'${script} ${entry} ${target}'`;
-		}).join(' ');
+		concurrentComands = concurrentComands.length ? concurrentComands : Object.values(commands);
```
leggsimon: Using `Object.values()` will only work if you're using > Node 7.5.

build/app/cli.js
```diff
@@ -54,16 +54,29 @@ program
...
+			'sass-only': cssEntryPoints.map(([target, entry]) => `'${script} ${entry} ${target}'`).join(' ')
+		};
+
+		for(let key in commands) {
+			if(!!options[key]) {
+				concurrentCommands.push(commands[key]);
+			}
+		}
+
+		concurrentCommands = concurrentCommands.length ? concurrentCommands : Object.values(commands);
```
leggsimon: `Object.values()` will only work for users on > Node 7.5 😢

PR #1099, opened 2017-10-06
Added 'nui compress' command

adambraimbridge:
# TODO: Move this PR to [n-heroku-tools](https://github.com/Financial-Times/n-heroku-tools)

Added `nui compress` command: 

> Generate compressed copies of any assets (css, js, map, json etc) in the `/public` directory.

This is not currently used by anything, but I plan to wire it up in the `nui deploy` pipeline.

image
For the `next-frontpage` app, its production build of `/public` assets is compressed from 1.9mb to 291kb. 

image
The two core `/public/n-ui` files are also compressed.

image
Here's how it looks when running `nui compress` in `next-front-page`

image
The compressed files are saved to their own `./compressed` directory.

### DIFF
--- a/build/lib/generate-compressed-assets.js
+++ b/build/lib/generate-compressed-assets.js
@@ -1,0 +1,35 @@
+/* eslint-disable no-console */
+const fs = require('fs');
+const path = require('path');
+const denodeify = require('denodeify');
+const compress = denodeify(require('iltorb').compress);
+const readFile = denodeify(fs.readFile);
+const writeFile = denodeify(fs.writeFile);
+
+const read = dir => fs.readdirSync(dir).reduce((files, name) => {
+	if (fs.statSync(path.join(dir, name)).isDirectory()) {
+		return files.concat(read(path.join(dir, name)));
+	} else {
+		return files.concat({
+			uncompressed: {dir, name},
+			compressed: {dir: `${dir}/compressed/`, name: `${name}.br`}
+		});
+	}
+}, []);
+
+module.exports = assetPath => {
+	const files = read(assetPath);
+	return Promise.all(files.map(file => {
+		if (!fs.existsSync(file.compressed.dir)) {
+			fs.mkdirSync(file.compressed.dir);
+		}
+		return readFile(path.join(file.uncompressed.dir,file.uncompressed.name))
+			.then(buffer => compress(buffer,{quality: 11}))
+			.then(contents => {
+				return writeFile(path.join(file.compressed.dir,file.compressed.name), contents)
+					.then(() => path.join(file.compressed.dir,file.compressed.name))
+					.catch(error => console.error(error));
+			});
+		}
+	)).catch(error => console.error(error));
+};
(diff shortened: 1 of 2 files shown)


adambraimbridge:
... of course, the compressed file sizes are compared against uncompressed versions. To be completely fair, these `brotli` compressed files should be compared against `gzip` compressed equivalents.

wheresrhys:
couple of questions
- I'd expect to see some pink in this PR as well, given that your changes do something very similar to what was already there for the n-ui build, so there must be scope for reconciling the two and deleting some code. n-ui deploy already does brotlify files
- does this need to be a cli task? Is there a use case for being able to do this outside of deploy-hashed-assets? Should it live there instead? Should deploy-hashed-assets be migrated into n-ui

build/lib/generate-compressed-assets.js
```diff
@@ -0,0 +1,35 @@
...
+const fs = require('fs');
+const path = require('path');
+const denodeify = require('denodeify');
+const compress = denodeify(require('iltorb').compress);
+const readFile = denodeify(fs.readFile);
+const writeFile = denodeify(fs.writeFile);
+
+const read = dir => fs.readdirSync(dir).reduce((files, name) => {
+	if (fs.statSync(path.join(dir, name)).isDirectory()) {
+		return files.concat(read(path.join(dir, name)));
```
adambraimbridge: ... maybe this should be 
```diff
-		return files.concat(read(path.join(dir, name)));
+		return read(path.join(dir, name));
```

adambraimbridge:
Thanks @wheresrhys 

> I'd expect to see some pink in this PR as well, given that your changes do something very similar to what was already there for the n-ui build, so there must be scope for reconciling the two and deleting some code. n-ui deploy already does brotlify files

Yeah. I can't understand fully how that already works, whether it's completed, and what needs to be done to get it actually loading in apps. I know there's a PR pending to serve the brotlified version (`n-ui-core.css.br`) via Fastly, but  from what what I can see, apps (like front page, streams and articles) don't actually load the uncompressed `n-ui-core.css` file. I'm not saying it doesn't — just that I can't figure it out.

> does this need to be a cli task? 

It doesn't _have_ to be a cli task. But because I can't really understand `n-ui` well enough to refactor/update the existing code, I needed to keep this `compress` functionality separate from the other functions. And the only way I could get _that_ to work (at least for development purposes) was to make it a cli task. This way I can develop and test `compress` independent of `build` and `deploy`.

> Is there a use case for being able to do this outside of deploy-hashed-assets? Should it live there instead? Should deploy-hashed-assets be migrated into n-ui?

I have no idea. I thought `nht` is deprecated? Even if it isn't, I'm really confused what the business logic is to decide which functionality belongs in which of our build tools. 

Now that we have a PR for reference, shall we set aside a half hour to help me understand all this, and then decide what the next steps are?

Cheers </a>

adambraimbridge:
Closing this PR in favour of trying an s3-auto-brotlify-lambda approach.

PR #1084, opened 2017-09-07
Webpack 3

remybach:
I've pointed a number of local apps to the test release I created for this and everything seemed to build and run just fine.

PR #1100, opened 2017-10-09
bump dependencies and update tests

doramatadora:
Suspect (cf previous bug duty) that https://sentry.io/nextftcom/ft-next-frontend/issues/316532770/events/ is webpack-related > https://github.com/webpack/webpack/issues/4916

Bumps webpack and dependencies. Incorporates https://github.com/Financial-Times/n-ui/pull/1084

I need some clarification for `loader.js` - is `broadcast` called after `console.error`?
 🐿 v2.5.16


package.json
```diff
@@ -26,86 +26,84 @@
...
-    "babel-core": "^6.25.0",
-    "babel-loader": "^7.1.1",
+    "autoprefixer": "^7.1.5",
+    "aws-sdk": "^2.130.0",
+    "babel-core": "^6.26.0",
+    "babel-loader": "^7.1.2",
     "babel-plugin-add-module-exports": "^0.2.1",
-    "babel-plugin-transform-es2015-classes": "^6.14.0",
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.8.0",
+    "babel-plugin-transform-es2015-classes": "^6.24.1",
```
wheresrhys: can these two deps be removed now we're on preset-env?

doramatadora: @wheresrhys `modules-commonjs` can be removed (unused). Provided there is a way to configure `transform-es2015-classes` to `loose` in the preset flags, I'll remove `babel-plugin-transform-es2015-classes` too. Agreed on the beta release.

i-like-robots: re. loose mode - as IE10 should now be considered a core browser I assume we can drop it? https://babeljs.io/docs/usage/caveats/#internet-explorer-classes-10-and-below-

doramatadora: thanks, @i-like-robots – I'm for dropping it

remybach:
Further to @wheresrhys's point, when I was working on this, you could make a pre-release that you test live on the `next-feedback` app as it has a fairly low amount of traffic.

doramatadora:
@remybach good suggestion. However, we won't know from `next-feedback` whether this fixes the errant Promise shim: https://sentry.io/nextftcom/ft-next-frontend/issues/316532770/events/

Makefile
```diff
@@ -128,7 +128,11 @@ deploy-s3:
 	# deploy to hashed urls on s3
 	nht deploy-hashed-assets --directory public/n-ui --monitor-assets
 
-deploy: deploy-s3 npm-publish rebuild-user-facing-apps
+publish-npm:
```
wheresrhys: can this just go into n-gage? maybe make the npm-publish task run this code if it detects the installed npm supports these options, otherwise fallback to the old version

PR #1105, opened 2017-10-11
Add headline uuid to page view event

missamynicholson:
🐿 v2.5.16


missamynicholson:
@tom-parker 
I was going to use the `data-content-id` attribute, but teaser images don't have this.
So can either add that property to all teasers, or can use the href like this.

tom-parker:
👍

PR #1106, opened 2017-10-16
Lazy loading optimisation fix - app names were wrong

alexflorisca:
🐿 v2.5.16


alexflorisca:
App names have since changed
`front` => `front-page` and 
`stream` => `stream-page`

And the optimisation for loading ads before they come into view stoped working.

PR #1108, opened 2017-10-17
polyfill es6 promise synchronously via babel

doramatadora:
🐿 v2.5.16

### DIFF
--- a/build/webpack.common.config.js
+++ b/build/webpack.common.config.js
@@ -65,6 +65,8 @@ module.exports = {
 					plugins: [
 						// converts `export default 'foo'` to `exports.default = 'foo'`
 						require.resolve('babel-plugin-add-module-exports'),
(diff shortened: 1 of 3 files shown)


build/webpack.common.config.js
```diff
@@ -65,7 +65,8 @@ module.exports = {
 					plugins: [
 						// converts `export default 'foo'` to `exports.default = 'foo'`
 						require.resolve('babel-plugin-add-module-exports'),
-
+						// rewrites Promise references to es6-promise, but only if necessary
```
wheresrhys: could you add a `//HACK comment` as we really shoudl get to the bottom of why this is happening

doramatadora:
Not needed - we error handle dynamic script loading now: https://github.com/Financial-Times/n-ui/pull/1111

PR #1111, opened 2017-10-19
Dora/v7 error handling

doramatadora:


browser/layout/partials/bootstrapper.html
```diff
@@ -8,19 +8,30 @@
...
-			var s = d.getElementsByTagName('script')[0];
-			s.parentNode.insertBefore(o, s.nextSibling);
-			return o;
+		const loadError = function (e) {
+			window.FT.core = true;
+			throw new URIError('Failed to load ' + e.target.src + ': falling back to core experience.');
+		};
+
+		const loadScript = function (src, isFatal) {
+			if(!window.FT.core) {
```
wheresrhys: This is too late to respond to the error as all the calls to `loadScript()` happen synchronously straight after each other, so even if one of them errored on the next tick of the event loop (which it never would as, I figure, most of the time failed loads are due to timeouts or dropped packets, which take a while to happen), it woudl still be too late to have any effect in here.

But even without this check in here, what you've done is useful as throwing the error will cause all scripts to stop executing. We can merge this into v7 and see if it has the expected effect of converting all those odd sentry errors into the one you're throwing above. If it does then we can look at putting something in our o-errors config to silence them/send to spoor instead

wheresrhys: so the feedback is remove the window.FT.core check as it won't have the effect you're aiming for

browser/layout/partials/bootstrapper.html
```diff
@@ -8,19 +8,26 @@
...
+			throw new URIError('Failed to load ' + e.target.src + ': falling back to core experience.');
+		};
+
+		const loadScript = function (src, stopsExecutionOnLoadError) {
 			var d = document, o = d.createElement('script');
 			{{! For dynamically created scripts this has the same effect as adding the defer
 				attribute to inline scripts, i.e. load asynchronously, execute in order}}
 			o.async = false;
 			o.src = src;
+			if(stopsExecutionOnLoadError) {
```
wheresrhys: :+1:

PR #1113, opened 2017-10-19
turn off o-errors when a script fails to load

wheresrhys:
🐿 v2.5.16

### DIFF
--- a/browser/bundles/o-errors.js
+++ b/browser/bundles/o-errors.js
@@ -1,11 +1,3 @@
-// for safety polyfill window.console
-window.console = window.console || {
-	log: function () { },
-	info: function () { },
-	warn: function () { },
-	error: function () { },
-};
-
 const oErrors = require('o-errors');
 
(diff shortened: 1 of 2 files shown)


browser/bundles/o-errors.js
```diff
@@ -1,11 +1,3 @@
-// for safety polyfill window.console
```
wheresrhys: safety from what though? http://caniuse.com/#search=console

wheresrhys:
depends on this https://github.com/Financial-Times/o-errors/pull/49

wheresrhys:
I also have an idea that n-ui should export a method - `onBeforeAppInitialize()`, or maybe `checkEnv()` - which throws an error, exiting all js, if the page is in an error state. Will lead to a more consistent user experience I think. Just an idle thought for now, and we could add it later as it will never be compulsory, rather a 'apps should call this method before attempting to execute'

PR #1119, opened 2017-11-03
Preconnect third party urls

alexflorisca:
🐿 v2.5.16

### DIFF
--- a/browser/layout/wrapper.html
+++ b/browser/layout/wrapper.html
@@ -17,11 +17,15 @@
 		{{/each}}
 		<link rel="preconnect" href="https://spoor-api.ft.com">
 		<link rel="preconnect" href="https://session-next.ft.com" crossorigin="use-credentials">
-
+		<link rel="preconnect" href="https://cdn.krxd.net">
+		<link rel="preconnect" href="https://beacon.krxd.net">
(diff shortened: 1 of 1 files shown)


browser/layout/wrapper.html
```diff
@@ -17,11 +17,15 @@
 		{{/each}}
 		<link rel="preconnect" href="https://spoor-api.ft.com">
 		<link rel="preconnect" href="https://session-next.ft.com" crossorigin="use-credentials">
-
+		<link rel="preconnect" href="https://cdn.krxd.net">
+		<link rel="preconnect" href="https://beacon.krxd.net">
 		<link rel="preconnect" href="https://ads-api.ft.com">
```
adgad: I'd argue Krux shouldn't be here...IIRC we don't don't _depend_ on Krux to render ads (it's done after), and so I assume it's better not to hog up the network?

alexflorisca: hmm I think it's supposed to have the opposite effect no? It gives extra information for the browser to optimise the request later on? From reading around, I think the browser is pretty good at optimising those prefetchs.

adgad: Try it and see 😬

PR #1125, opened 2017-11-07
Filter out window.FT errors

alexflorisca:
🐿 v2.5.16

### DIFF
--- a/browser/bundles/o-errors.js
+++ b/browser/bundles/o-errors.js
@@ -14,8 +14,8 @@ oErrors.init({
 	tags: {
 		appName: appInfo.name
 	},
-	filterError: function () {
-		return !window.FT.disableOErrors;
+	filterError: function (reportedObject) {
+		return !(window.FT.disableOErrors || reportedObject.error.toString().search('window.FT') > 0);
 	},
 	errorBuffer: window.errorBuffer || []
 });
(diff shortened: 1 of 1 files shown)


browser/bundles/o-errors.js
```diff
@@ -14,8 +14,8 @@ oErrors.init({
 	tags: {
 		appName: appInfo.name
 	},
-	filterError: function () {
-		return !window.FT.disableOErrors;
+	filterError: function (reportedObject) {
+		return !(window.FT.disableOErrors || reportedObject.error.toString().search('window.FT') > 0);
```
wheresrhys: maybe that's a little too loose. Would it be better to check for `window.FT.flags || window.FT.nUi`?

alexflorisca:
@wheresrhys How about now? I think there's more places where that errors shows up. I've seen it in `.flags`, `.nUi`, `.ftNextUi` and possibly others.

wheresrhys:
It *shoudl* just be those three values though. Would prefer to scope it more strictly for now as we don't want it to swallow errors we might not have considered and that might be _real_

alexflorisca:
@wheresrhys Ok how about now?

wheresrhys:
cool ta

sjparkinson:
image

👏

PR #1129, opened 2017-11-09
Update to latest n-desktop-app-banner.

remybach:
The subscriberCohort

 🐿 v2.5.16


tavvy:
i thought we only wanted to show to subscribers as you must be a subscriber to use the app?

remybach:
Yep, confirmed with @doramatadora... the brain already makes that check so it's currently an unnecessary extra check.

PR #1131, opened 2017-11-13
Turn off lazy loading ads on homepage for Q4

alexflorisca:
🐿 v2.5.16


wheresrhys:
My initial thoughts are that ad ops need to really justify this and should'n't just assume they can disable a perf optimisation. Where's the data to justify that there's a problem? Why are they asking us to implement a solution rather than explaining the problem and asking us to solve it?

bjfletcher:
@alexflorisca We'll have a chat about this. Hold fire for the time being.

alexflorisca:
For posterity - we decided to do this, but using a separate flag `noLazyLoadingFrontPage` which expires in January.

PR #1140, opened 2018-01-02
make the hardcoded current year in the footer dynamic

bjfletcher:
🐿 v2.6.0

### DIFF
--- a/server/models/current-year.js
+++ b/server/models/current-year.js
@@ -1,0 +1,4 @@
+module.exports = (req, res, next) => {
+	res.locals.currentYear = new Date().getFullYear();
+	next();
+};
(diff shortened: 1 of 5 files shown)


server/models/current-year.js
```diff
@@ -0,0 +1,4 @@
+module.exports = (req, res, next) => {
+	res.locals.currentYear = new Date().getFullYear();
```
leggsimon: Let's hope the servers have the right time set up 🤞

bjfletcher: Heh :) the email services (and apparently myFT) have been using this for a while and so there's prior proven use ☺️

bjfletcher:
Screengrab of what it looks like locally with the PR: image

PR #1141, opened 2018-01-02
add Speedcurve LUX

bjfletcher:
🐿 v2.6.0

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,2 @@
+exclude_patterns:
+- "components/n-ui/tracking/third-party/speedcurve-lux/inline-code.js"
--- a/components/n-ui/tracking/index.js
+++ b/components/n-ui/tracking/index.js
@@ -4,6 +4,7 @@ module.exports = {
 	},
 	lazyInit: function (flags) {
 		window.addEventListener('ftNextLoaded', function () {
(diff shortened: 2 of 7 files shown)


wheresrhys:
On the previous attempt to add it (when we tried the alpha) there was also an inline script. https://github.com/Financial-Times/n-ui/pull/377/files 

Is that not needed these days?

bjfletcher:
@wheresrhys wow that's some mad inline code in the early days... yup, seems like they've since improved

https://speedcurve.com/features/lux/#snippet says just the LUX config object is needed.

bjfletcher:
@wheresrhys ok, I think I have covered everything...
put in inline code from Speedcurve for mark/measure functions
this inline code is in a file of its own, meaning it's easier for me to get the linters/coveralls to ignore this file (it breaks many rules!)

wheresrhys:
Isn't the point with the inline code though that it has to be inlined in the html?

bjfletcher:
@wheresrhys hopefully no - if it doesn't work, we can figure out how to set it up so that it gets run sooner

bjfletcher:
(unless you think we should put the JS directly into the HTML, although that's a bit messy IMHO)

sjparkinson:
Closing because #1142 was merged.

PR #1142, opened 2018-01-04
add Speedcurve LUX the HTML way

bjfletcher:
rather than the JS way


bjfletcher:
@wheresrhys I think you wanted it this way... happy for you to choose: HTML (this PR) or JS (the other PR)

adgad:
Are we running this for all our users? Is it better to sample perhaps?

browser/layout/vanilla.html
```diff
@@ -6,6 +6,9 @@
 		<title>{{#title}}{{{this}}}{{/title}}</title>
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta name="theme-color" content="#fff1e0">
+		{{#unless @root.flags.speedcurveLux}}
```
remybach: Shouldn't this be `if`? I know @leggsimon has strong feelings about using a flag in this way.

remybach:
Oops! Only meant to comment, not necessarily approve it... apologies.

browser/layout/vanilla.html
sjparkinson: Correct! The snippet below is to unregister the service worker if the flag is off, so `unless` makes sense there.

sjparkinson:
> Set the sample rate to a value that matches the LUX plan you want to sign up for. For example, if your site typically gets 900 million page views per month, and you want one LUX plan that covers a maximum of 10 million page views per month, then set the sample rate to LUX.samplerate=1; (which is 1%).

From https://speedcurve.com/features/lux/#snippet.

browser/layout/vanilla.html
leggsimon: I'm glad you're thinking of me @remybach 🙃

sjparkinson:
> Set the sample rate to a value that matches the LUX plan you want to sign up for. For example, if your site typically gets 900 million page views per month, and you want one LUX plan that covers a maximum of 10 million page views per month, then set the sample rate to LUX.samplerate=1; (which is 1%).

From https://speedcurve.com/features/lux/#snippet.

But from Fastly, in December we had 5.5 million page views.

browser/layout/vanilla.html
bjfletcher: @remybach @leggsimon do you know how to negate in Handlebars? If we can, then we don't need `unless` (unless we want to use the empty `if` with it all in the `else` block)

bjfletcher:
@adgad options for launch:
1. A/B flag this... so `speedcurveLux` is turned on for only certain % of users, or perhaps a certain cohort...
2. leave it off by default, only us developers can turn it on (via toggler.ft.com) and see how it behaves with the dashboards, etc...
3. go the full throttle and get the full data, for all users, for the next week... can turn it off again if we find issues

browser/layout/vanilla.html
leggsimon: IIC you can't negate in normal handlebars, `n-handlebars` adds the unless helper method

adgad:
@bjfletcher For launcher I'm fine with either to get some data (well, 1 or 3 will get us better data). 

Mid term was just wondering if it's worth not serving the JS to all users. Using the flag to do that seems like a good idea.

browser/layout/vanilla.html
bjfletcher: :boom: what were I thinking... 0 out of 10 for coding effort, Ben! Thanks Remy & Simon ❤️

bjfletcher:
@adgad they're pretty good, I've done some testing here... without LUX, total download for n-ui's demo page is 1.2mb, with LUX, it's still 1.2mb...
In fact, I think it's about 10kb... the script we load from them is only 10kb including headers

PR #1146, opened 2018-01-10
Some stylelint happiness

benbarnett:
Fixes the following:

screenshot 2018-01-10 16 03 03

🐿 v2.5.16


leggsimon:
The `!important`s aren't failing it though are they? They just look like warning, looks like the `selector-no-qualifiing-type` one is the one that's failing the build?

benbarnett:
@leggsimon it is a point though, I hadn't realised I didn't need to until just now..

PR #1148, opened 2018-01-11
a safer way to stringify error object

bjfletcher:
Big thanks to @JakeChampion for helping to make this more robust.

### DIFF
--- a/browser/bundles/o-errors.js
+++ b/browser/bundles/o-errors.js
@@ -16,7 +16,7 @@ oErrors.init({
 	},
 	filterError: function (reportedObject) {
 		// Supress errors if they contain an "Undefined not a.... evaluating window.FT.flags | .nUi | .ftNextUi type error"
-		const e = reportedObject.error.toString();
+		const e = Object.prototype.toString.call(reportedObject.error);
 		const windowFtError = !!e.match(/^(?=.*\bundefined\b)(?=.*(\bwindow.FT.flags\b|\bwindow.FT.nUi\b|\bwindow.FT.ftNextUi\b)).*$/gi);
 		return !(window.FT.disableOErrors || windowFtError);
 	},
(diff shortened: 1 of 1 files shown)


browser/bundles/o-errors.js
```diff
@@ -16,7 +16,7 @@ oErrors.init({
 	},
 	filterError: function (reportedObject) {
 		// Supress errors if they contain an "Undefined not a.... evaluating window.FT.flags | .nUi | .ftNextUi type error"
-		const e = reportedObject.error.toString();
+		const e = Object.prototype.toString.call(reportedObject.error);
```
JakeChampion: My suggested fix is wrong, sorry.
You want `const e = String(reportedObject.error);`
`Object.prototype.toString.call(reportedObject.error);` will return the name of the object type and not the contents of the object.
E.G.
```> Object.prototype.toString.call([1,2,3])
'[object Array]'
> String([1,2,3])
'1,2,3'```

Comment (bjfletcher) [reply browser/bundles/o-errors.js]: I'm not sure what's the intention of this code

The old way - `reportedObject.error.toString()` would return the `[object Array]` anyway, right?

maybe @wheresrhys can clarify

Comment (bjfletcher) [reply browser/bundles/o-errors.js]: for context: https://github.com/Financial-Times/n-ui/pull/1125

Comment (JakeChampion) [reply browser/bundles/o-errors.js]: It would not, it would have called `Array.prototype.toString` which would return `'1,2,3'`. 

However not all objects implement a `toString` method E.G. `Object.create(null);`, this would also not work with the string constructor approach...

The best solution I can come up with is 
```js
let e;
if ('error' in reportedObject) {
    if ('toString' in reportedObject.error && typeof reportedObject.error.toString === 'function') {
        e = reportedObject.error.toString();
    } else {
        try {
            e = String(reportedObject.error);
        } catch (err) {
            e = Object.prototype.toString(reportedObject.error);
        }
    }
}
```

bjfletcher: image

yup, et's go with your new suggestion :)

JakeChampion: `String(Object.create(null))` throws an error, which is why my latest code block contained the `catch`. You could simplify my latest code block to this:
```js
let e;
if ('error' in reportedObject) {
    try {
        e = String(reportedObject.error);
    } catch (err) {
        /* 
            This means the object could not be converted to a primitive value.
            This wold happen for `Object.create(null)`.
        */
    }
}
```

PR #1149, opened 2018-01-12
get the o-errors stuff tested

bjfletcher:
so we can define what behaviour we want to see... you know, BDD and all that ;)


bjfletcher:
@doramatadora @tavvy I'll merge this... feel free to add more tests to ensure we're filtering the right things

PR #1150, opened 2018-01-15
Dora/typescript

doramatadora:


circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 8.6.0
+    version: 8.9.3
```
jonnywyatt: 8.9.4 is the latest LTS version?

doramatadora: Indeed it is!

PR #1158, opened 2018-01-19
Fix: Allow node-sass and postcss to handle different filenames.

adambraimbridge:
I'd like to get this fix out ASAP. Then I can look into how this bug managed to pass all the tests.

### DIFF
--- a/scripts/build-sass.sh
+++ b/scripts/build-sass.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


scripts/build-sass.sh
```diff
@@ -1,15 +1,20 @@
-#!/bin/sh
+#!/bin/bash
```
adambraimbridge: I need to learn why `#!/bin/bash` works but `#!/bin/sh` errors. See: https://stackoverflow.com/a/23277604/2194699

PR #1160, opened 2018-01-22
Add all Ammit tests to LUX

bjfletcher:
This would allow us to compare performance impact between all variants of all running Ammit tests.


adgad:
Nice! Is there any downside on Speedcurve's end to sending _all_ the flags? Would it be worth having a property on the flag (e.g. performance: true) or something like that?

missamynicholson:
These are just the AB test flags, so there shouldn't be too many per page.

bjfletcher:
Thanks @adgad – yup, @missamynicholson is right... only on active Ammit tests
Speedcurve encourages things like this and it's why we pay them a load of $... so they can do it ;)

PR #1162, opened 2018-01-23
don't track sw flags if sw not supported

wheresrhys:
Because otherwise our mobile data is swamped by iOS and difficult to detect the size of the effect.


bjfletcher:
@wheresrhys cool

PR #1174, opened 2018-02-07
client-side typescript transpile

doramatadora:


build/webpack.common.config.js
```diff
@@ -52,43 +80,21 @@ module.exports = {
...
 			{
-				test: /\.ts?$/, // Another convention is to use the .es6 filetype, but you then
-				// have to supply that explicitly in import statements, which isn't cool.
+				test: /\.ts?$/,
 				exclude: [/(node_modules|bower_components)/],
-				loader: 'ts-loader'
+				use: [
+					babelLoaderConfig,
+					{
+						loader: 'ts-loader'
```
i-like-robots: Because TS is based on Babel do we potentially have a double-Babel situation going on here... Is the intention for TS loader here to output `esnext` code (i.e. just strip the TS annotations) and the inherited Babel loader to do the rest? And if so, should any expected`tsconfig.json` rules be defined here?

doramatadora: According to the `ts-loader` docs, there is no danger: [example usage with babel](https://github.com/TypeStrong/ts-loader/blob/master/examples/react-babel-karma-gulp/webpack.config.base.js)

`n-ui` need be opinionated only on the client-side js it bundles, but we could have a common `tsconfig.json` in `n-gage`.

Makefile
```diff
@@ -59,10 +59,10 @@ else
 endif
 
 test-server-plain:
-	mocha server/test/*.test.js server/test/**/*.test.js
+	mocha --exit server/test/*.test.js server/test/**/*.test.js
```
kavanagh: Why did you need to add this?

doramatadora: @kavanagh The default behaviour of `mocha ^4` is not to exit when tests finish running (unlike mocha 3). I'd initially bumped mocha as part of a mass dependency bump, but `karma-mocha` doesn't support passing the exit flag, so I reverted to mocha 3. 

Forgot about the flag in Makefile – it doesn't do anything with mocha 3. Shall I remove it or leave it there for a future bump (once `karma-mocha` supports correct behaviour)?

kavanagh: Good detail, thanks! I think leave it as is.

PR #1175, opened 2018-02-07
Pass ammit allocations with the event context

tom-parker:
🐿 v2.7.0

The ammit allocations for a given user exist on the `data-ab-state` attribute of the html tag. This change passes an `active_ammit_flags` object into the o-tracking context so the ammit data is sent with every event. 

@tbergmen @missamynicholson

### DIFF
--- a/components/n-ui/tracking/ft/index.js
+++ b/components/n-ui/tracking/ft/index.js
@@ -98,6 +98,16 @@ const oTrackingWrapper = {
 				}
 			}
 
+			const abState = getRootData('ab-state');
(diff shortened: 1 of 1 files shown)


missamynicholson:
:+1:

tom-parker:
@missamynicholson @tbergmen I've since remembered if the call to ammit fails and `data-ab-state="-"` we actually want to pass that with the events so we don't incorrectly enrich things

missamynicholson:
Good point!

PR #1182, opened 2018-02-12
add colors dependency

doramatadora:
how did we [get away](https://github.com/Financial-Times/n-ui/blob/master/build/app/logger.js#L1) without it?


adgad:
👍

PR #1184, opened 2018-02-20
Don't show the marketing promo banner for My Account pages

jonnywyatt:
🐿 v2.5.16

### DIFF
--- a/components/n-ui/header/js/promoHandler.js
+++ b/components/n-ui/header/js/promoHandler.js
@@ -34,6 +34,10 @@ function isSignupForm () {
 		|| window.location.pathname === '/thank-you';
 }
 
+function isMyAccount () {
+	return document.documentElement.dataset && document.documentElement.dataset.nextApp === 'control-centre';
+}
+
 function isBarrier () {
 	return !!document.querySelector('.barrier');
 }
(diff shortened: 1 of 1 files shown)


doramatadora:
nice solution. It's definitely `control-centre` and not `next-control-centre`, right?

doramatadora:
I suggest using this approach for barriers and signup, too – less brittle than checking for classes in the markup, and quicker. @tavvy @alexnaish

tavvy:
👍  we will fix that and more when VoltDB is up in the US https://github.com/Financial-Times/n-messaging-client/issues/31 🎉

components/n-ui/header/js/promoHandler.js
```diff
@@ -34,6 +34,10 @@ function isSignupForm () {
 		|| window.location.pathname === '/thank-you';
 }
 
+function isMyAccount () {
+	return document.documentElement.dataset.nextApp === 'control-centre';
```
leggsimon: Just in case, it's probably worth ensuring `document.documentElement.dataset` exists before querying `nextApp` on it.

jonnywyatt: done, thanks

PR #1185, opened 2018-02-21
Slightly more optimal CTM

adgad:
Very minor optimisation, but constructing a blob is up to 1000 times slower than `typeof Blob (https://jsperf.com/blobs), and evaluating this sometimes takes up to 40ms.

screen shot 2018-02-21 at 16 08 58

(note: need to check that Blob isn't still a function in Android 4 - which is what this test was designed to exclude)

### DIFF
--- a/browser/layout/partials/ctm.html
+++ b/browser/layout/partials/ctm.html
@@ -2,17 +2,15 @@
 var classes = document.documentElement.className; {{!note: once we ditch support for ie9 we can just use classlist}}
 classes = classes.replace(/\bno-js\b/, 'js'); {{!note: needed for JS that's run out of the scope of the bootstrap (inline onclicks, etc)}}
 
-var supportsBlobConstructor = typeof Blob === 'function';
 var script = document.createElement('script');
 var supportsDeferredScripts = "defer" in script && "async" in script;
 
 window.cutsTheMustard = (
-	typeof Function.prototype.bind !== 'undefined'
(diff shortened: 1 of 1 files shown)


i-like-robots:
-1000%

leggsimon:
Hmm, I'm not sure whether this will work. I seem to remember that `Blob()` works in Android but `new Blob` doesn't. That was why it was in the try/catch, so that it would catch the error of calling new.

adgad:
You are correct. New proposal:

https://caniuse.com/#feat=pagevisibility

`'visibilityState' in document`

adgad:
Pushed a simpler CTM. Tested on browsers below (tick = CTM, no tick = doesn't CTM)


- [ ] Android 4.4 Browser 
- [ ] Android 4.4 Chrome
- [ ] iOS Safari 10 - 
- [ ] Edge
- [ ] Chrome (latest)
- [ ] Firefox (latest + 48)
- [ ] Desktop Safari(11 and 9)
- [ ] IE11
- [ ] IE10

adgad:
https://jsperf.com/ft-ctm comparison of the two

adambraimbridge:
image
Whoa

leggsimon:
This looks like it'll still pass for Android 4.4 & 4.4.4? Is that right?

adgad:
Nope - visibilityState is prefixed with `webkit` in android 4.4

PR #1186, opened 2018-02-22
Remove the nCounterAdBlocking code

adambraimbridge:
Pending results of an A/B test that will hopefully show that nCounterAdBlocking (specifically, sourcepoint) makes performance worse, for no measurable benefit.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,7 +19,6 @@
     "fontfaceobserver": "^2.0.9",
     "ftdomdelegate": "^2.0.3",
     "loadcss": "filamentgroup/loadCSS#^1.2.0",
-    "n-counter-ad-blocking": "^3.1.2",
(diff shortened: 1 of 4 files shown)


adgad:
I think sourcepoint might be somewhere else in n-ui. Tracking maybe?

adambraimbridge:
From what I can gather, the sourcepoint script is added via `n-counter-ad-blocking`, here: https://github.com/Financial-Times/n-counter-ad-blocking/blob/master/main.js#L35

This PR removes `n-counter-ad-blocking` from `n-ui`. If it were merged, the cleanup task may well include deleting `n-counter-ad-blocking` entirely — unless some other app is using it ...

adambraimbridge:
This has been done separately. 👍

PR #1188, opened 2018-02-22
Removed logic for 'nUiExcludes' because nothing uses it

adambraimbridge:
image

I can't find any example of a `nUiExcludes` property, so I'm simplifying the code

### DIFF
--- a/browser/test/n-ui-bundle.spec.js
+++ b/browser/test/n-ui-bundle.spec.js
@@ -50,7 +50,7 @@ describe('n-ui bundle', () => {
 		});
 
 	describe('_entry', () => {
-		const aliases = entry();
+		const aliases = entry;
 
 		const getProp = str => {
 			const nest = str.split('.');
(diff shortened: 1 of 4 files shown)


adgad:
Out of curiosity, what does nUiExcludes actually do?

leggsimon:
If I remember correctly, nUiExternals is a way you can define an "external" usage of a module. Normally, when n-ui builds an app’s JS if it sees a `require('o-date')` for example it'll strip it out and use `window.FT.nUi._hiddenComponents.oDate` instead. This is so that, for some modules we only have one version knocking about but if for whatever reason an app needed to use a specific version of o-date it could define that it is "external".

I'm not sure whether we want to get rid of this option...

adambraimbridge:
Thinking about it now, if we could understand the actual problem that is solved by `nUiExternals` (aka `webpack-externals)`,  we'd be in a better position to judge whether the solution is still required, and whether we can improve upon it. And of course document it a little better.

leggsimon:
🤦‍♂️ I just realised I didn't answer your question @adgad. I'm almost certain there was a reason for having it but I can't remember what it was. @wheresrhys may remember...

PR #1189, opened 2018-02-23
run webfonts loader script only when necessary, and put it behind a flag

bjfletcher:
webfonts loader script:
1. put it behind a flag
2. include it only if fonts aren't already present


edds:
This PR is being closed due to inactivity as part of a Pull Request Bankruptcy. If you feel it’s still useful or relevant please feel free to re-open it.

PR #1194, opened 2018-02-28
Quickly make the background color of the page pink

kavanagh:
In case the browser is having trouble rendering anything we might at least be able to show something vaguely FT is happening. It might also help us learn more about the white-screen-of-doom (maybe).


i-like-robots:
👍

benbarnett:
Makes me miss the `<body bgcolor="pink">` days

kavanagh:
@benbarnett we could do that for old times sake and annoy/amuse people looking out our code

PR #1197, opened 2018-03-02
Add some context.data.source info for each of the events.

remybach:
Two different `context.data.source` values will get sent:

1. `noscript`
2. `cuts-the-mustard-check`

I've also updated it to send the app name.


adgad:
Nice! 👍

PR #1200, opened 2018-03-09
Fixed filter errors

alexflorisca:
One of the keys from the callback argument object changed from `error` to `exception`. Who knew.


doramatadora:
@alexflorisca superstar!

alexflorisca:
Maybe we can filter out the service worker errors too? Couldn't work out why they only just started happening.

doramatadora:
@alexflorisca I'm reluctant to do so because I don't understand why the SW is erroring. We filter out script loading errors because they are neither preventable or fixable – I wonder if that's the case for SW.

PR #1206, opened 2018-03-20
Suppress service worker unavailable errors

remybach:


browser/js/app-initializer.js
```diff
@@ -104,7 +104,13 @@ export class AppInitializer {
 				.register(flags)
 				.catch(() => { });
 
-			serviceWorker.message({ type: 'updateCache', data: {}});
+			serviceWorker
+				.message({ type: 'updateCache', data: {}})
+				.catch((err) => {
+					if (err.toLowerCase() !== 'service worker unavailable') {
+						return Promise.reject(err);
```
wheresrhys: simpler to just `throw err`?

remybach: Ha... we had this discussion a few minutes ago. I just decided to do this since that's how it was being "thrown" in `n-service-worker`. Happy to change to `throw err` if it's preferred though. 👍

PR #1210, opened 2018-03-22
add postcss config to apply plugins options

fenglish:
postcss hasn't passed plugins(autoprefixer)'s option properly. I have added a postcss config file, and also set grid option in it.

Setting postcss config file is suggested in this issue.
https://github.com/postcss/postcss-cli/issues/144
 🐿 v2.8.0

### DIFF
--- a/scripts/postcss.config.js
+++ b/scripts/postcss.config.js
@@ -1,0 +1,8 @@
+module.exports = {
+	map: false,
+	plugins: [
+		require('postcss-discard-duplicates')(),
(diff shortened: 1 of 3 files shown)


scripts/postcss.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+	map: false,
+	plugins: [
+		require('postcss-discard-duplicates')(),
+		require('autoprefixer')({ browsers: '> 1%, last 2 versions, ie >= 9, ff ESR, bb >= 7, iOS >= 5', grid: true }),
```
wheresrhys: I woudl split these options onto multiple lines

fenglish:
@wheresrhys Thank you for your advise and review! I have added your advise in the trello ticket:+1:

PR #1211, opened 2018-03-22
Removed code for 'timeOnPage' event

adambraimbridge:
🐿 v2.8.0


benbarnett:
I think there's a spec file somewhere too @adambraimbridge 👍

adambraimbridge:
Thanks @benbarnett, so there is. Odd that removing the code hasn't failed the tests ... :)

benbarnett:
That is a point of the upmost excellence. Hmm!

adambraimbridge:
Hahahah :)

PR #1212, opened 2018-03-22
Fix (I think) make build-production breaking

adgad:
I'm guessing something weird with concurrently and quote marks.

Also, assuming that all the concurrent commands run in a spawned process, probably want to export CSS_SOURCE_MAPS in each of them? 🤷‍♂️

### DIFF
--- a/build/app/cli.js
+++ b/build/app/cli.js
@@ -65,8 +65,8 @@ program
 		const script = './node_modules/@financial-times/n-ui/scripts/build-sass.sh';
 		const commands = {
 			jsOnly: `'webpack --bail --config ${webpackConfPath} ${options.production ? '-p' : ''}'`,
-			sassOnly: `export CSS_SOURCE_MAPS=${!options.production} && ${cssEntryPoints
-				.map(([target, entry]) => `'${script} ${entry} ${target}'`)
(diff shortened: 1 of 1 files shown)


build/app/cli.js
```diff
@@ -65,8 +65,8 @@ program
 		const script = './node_modules/@financial-times/n-ui/scripts/build-sass.sh';
 		const commands = {
 			jsOnly: `'webpack --bail --config ${webpackConfPath} ${options.production ? '-p' : ''}'`,
-			sassOnly: `export CSS_SOURCE_MAPS=${!options.production} && ${cssEntryPoints
-				.map(([target, entry]) => `'${script} ${entry} ${target}'`)
+			sassOnly: `${cssEntryPoints
+				.map(([target, entry]) => `'export CSS_SOURCE_MAPS=${!options.production} && ${script} ${entry} ${target}'`)
```
adambraimbridge: By the way, this will `export CSS_SOURCE_MAPS` multiple times, right? One for each iteration in the `map()` ..?

adgad: Yup, _but_, if concurrently works the way I assume it would work (by spawning a new process for each command), I think it would have to be exported for each one anyway. (But that's a wild guess)

PR #1222, opened 2018-04-06
Allow header logo to not be a link

jonnywyatt:
For when page is linked to from iOS app and no nav should be available

 🐿 v2.8.0


leggsimon:
What is a use case for not having this as a link?

jonnywyatt:
Apparently, iOS app store approval is dependent on there being no outbound links from pages presented in its webview. Martin Fallon can confirm.

I [updated the readme](https://github.com/Financial-Times/n-ui/pull/1223) with explanation of these options

leggsimon:
TIL

PR #1225, opened 2018-04-09
add qualtrics tag code

leggsimon:
🐿 v2.7.0

### DIFF
--- a/browser/layout/partials/qualtrics-tag.html
+++ b/browser/layout/partials/qualtrics-tag.html
@@ -1,0 +1,63 @@
+<!--BEGIN QUALTRICS SITE INTERCEPT-->
+<script type='text/javascript'>
+	(function () {
+		var g = function (e, h, f, g) {
+			this.get = function (a) {
+				for (var a = a + "=", c = document.cookie.split(";"), b = 0, e = c.length; b < e; b++) {
+					for (var d = c[b];
+						" " == d.charAt(0);) d = d.substring(1, d.length);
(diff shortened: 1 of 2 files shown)


browser/layout/partials/qualtrics-tag.html
```diff
@@ -0,0 +1,63 @@
+<!--BEGIN QUALTRICS SITE INTERCEPT-->
```
adgad: Are there any docs/original hosting for this script that we can put in a comment?

browser/layout/partials/qualtrics-tag.html
```diff
@@ -0,0 +1,63 @@
...
+				if (this.check()) {
+					var a = document.createElement("script");
+					a.type = "text/javascript";
+					a.src = g + "&t=" + (new Date()).getTime();
+					document.body && document.body.appendChild(a)
+				}
+			};
+			this.start = function () {
+				var a = this;
+				window.addEventListener ? window.addEventListener("load", function () {
```
benbarnett: why not

benbarnett: have

benbarnett: 👇

adgad: Oh good point, this should probably be checking CTM

browser/layout/partials/qualtrics-tag.html
leggsimon: I'll add some later

PR #1227, opened 2018-04-10
Add nUi config option to hide footer

jonnywyatt:
To block all outbound links from the page if the user comes from iOS app, similar to recent [header changes](https://github.com/Financial-Times/n-ui/pull/1222#issuecomment-379693416)

 🐿 v2.8.0

### DIFF
--- a/browser/layout/wrapper.html
+++ b/browser/layout/wrapper.html
@@ -132,6 +132,7 @@
 				{{{body}}}
 			</div>
 
+			{{#unlessEquals nUi.footer.hide true}}
 			<div class="n-layout__row n-layout__row--footer">
 				{{#outputBlock 'footer'}}{{/outputBlock}}
 				{{>n-ui/footer/template}}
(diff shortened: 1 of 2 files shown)


browser/layout/wrapper.html
```diff
@@ -132,13 +132,15 @@
 				{{{body}}}
 			</div>
 
+			{{#unlessEquals nUi.footer.hide true}}
```
doramatadora: If this is always a boolean, `{{#unlessEquals nUi.footer.hide}}` would suffice.

jonnywyatt: it's true.. but I have seen in some codebases the use of 'false' as a string for example, I just wanted to make this strictly typed and bulletproof as it has a major impact

leggsimon: does footer have any other options?

leggsimon: It seems odd to have a `hide` property as an option.

jonnywyatt: I did consider just nUi.footer. I figured using the same structure as nUi.header made sense - less for a dev to consider, and leaves room for new options.
Open to alternative suggestions to hide property

leggsimon: I think we should mark this up to improve in v9

jonnywyatt: what's v9 sorry?

leggsimon: Sorry, the next major version. I added it to #1229

PR #1228, opened 2018-04-11
Add Fastly insights (RUM for CDN)

bjfletcher:
Fastly will monitor and ramp this up from 10% of the traffic up to 100%.

We can turn this on/off ourselves using the `fastlyInsights` feature flag.


sjparkinson:
Patrick Hamann is the contact at Fastly regarding this.

phamann@fastly.com

PR #1233, opened 2018-04-18
Added "evenMoreJanky" component (behind a flag)

adambraimbridge:
For a list of elements,
Immediately make them arbitrarily taller than their intended height.
Then over time, allow them to resume their intended height, thus creating a "janky" effect;
That is, the content "stutters" up/down the web page as the page layout updates.


simonplend:
Belatedly... :+1: Jank it to the max!

PR #1235, opened 2018-05-01
make consent available on resource locals

doramatadora:
**[GDPR] Profiling**

Bumps `n-express` and forces the `withConsent` app option:
If the `FT-Consent` header is present (from preflight), parse its content into `res.locals.consent`:

```
ft-consent=marketingByemail:on,recommendedcontentOnsite:off
```

--->

```
res.locals.consent = {
 	marketingByemail: true,
 	recommendedcontentOnsite: false
}
```

See Financial-Times/next-preflight#304 for background on the `FT-Consent` header.
See https://github.com/Financial-Times/n-express/pull/500 for `n-express` changes.

### DIFF
--- a/server/test/setup.js
+++ b/server/test/setup.js
@@ -1,0 +1,15 @@
+const sinon = require('sinon');
+const logger = require('@financial-times/n-logger').default;
+
+let loggerSandbox;
+
+beforeEach(() => {
+	loggerSandbox = sinon.sandbox.create();
+	[ 'info', 'warn', 'error' ].map(logLevel => {
+		loggerSandbox.stub(logger, logLevel);
+	});
+});
+
+afterEach(() => {
+	loggerSandbox.restore();
+});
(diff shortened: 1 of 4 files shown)


server/test/setup.js
```diff
@@ -0,0 +1,15 @@
+const sinon = require('sinon');
+const logger = require('@financial-times/n-logger').default;
+
+let loggerSandbox;
+
+beforeEach(() => {
+	loggerSandbox = sinon.sandbox.create();
+	[ 'info', 'warn', 'error' ].map(logLevel => {
+		loggerSandbox.stub(logger, logLevel);
```
adgad: Would this mask error logs for genuine test fails?

Perhaps it would be better to have the stub replace the call with a call to console.<logLevel>?

doramatadora: Excellent point. I'll change that.

PR #1241, opened 2018-05-04
Check if user has dismissed bottom cookie banner

debugwand:
🐿 v2.8.0


tavvy:
I dont think this will be needed if we bake the `FTCookieConsentGDPR` cookie check directly into o-cookie-message

Issue #1229, opened 2018-04-11
v9


leggsimon:
Ideas for the next major version

- [ ] Improve options for showing/hiding features ([header navigation links](), [footer]())

PR #1244, opened 2018-05-17
If hideOutboundLinks flag set, show only an unlinked logo in header

jonnywyatt:
🐿 v2.8.0


jonnywyatt:
I completely agree this should have a test. But N-ui has no test setup at all right now, and this change has to go live asap for GDPR compliance. I'll add a note to https://github.com/Financial-Times/n-ui/issues/1229

PR #1246, opened 2018-05-24
Bumping version of o-ads to include cookie consent check

andrewgeorgiou1981:
🐿 v2.8.0


doramatadora:
This should happen first: https://github.com/Financial-Times/o-ads/pull/128

PR #1248, opened 2018-05-24
Moat invalid traffic test

alexflorisca:
🐿 v2.8.0

### DIFF
--- a/browser/layout/wrapper.html
+++ b/browser/layout/wrapper.html
@@ -102,7 +102,9 @@
 				{{{json this}}}
 			</script>
 		{{/each}}
-		<script async id="moat-ivt" src="https://sejs.moatads.com/financialtimesprebidheader859796398452/yi.js"></script>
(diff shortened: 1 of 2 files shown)


browser/layout/wrapper.html
```diff
@@ -102,6 +102,9 @@
 				{{{json this}}}
 			</script>
 		{{/each}}
+		{{#if @root.flags.moatAdsTraffic}}
+			<script async id="moat-ivt" src="https://sejs.moatads.com/financialtimesprebidheader859796398452/yi.js"></script>
```
leggsimon: I noticed that this is still being requested even with the `ads` flag off. I think it should probably make sure the ads flag is also on

alexflorisca: @leggsimon Cheers for flagging, we're going to remove this flag shortly, just waiting on some tings.

PR #1267, opened 2018-06-12
Enable feedback in apps

dapenguin:


bower.json
```diff
@@ -33,6 +33,7 @@
     "o-footer": "^6.0.0",
     "o-header": "^7.5.6",
     "o-tracking": "^1.3.8",
-    "logo-images": "^1.5.2"
+    "logo-images": "^1.5.2",
+    "n-feedback": "^0.0.6"
```
kavanagh: It would be better to start at `1.0.0-beta.1`

browser/js/app-initializer.js
```diff
@@ -27,6 +28,7 @@ export const presets = {
 		ads: true,
 		syndication: true,
 		roe: true,
+		feedback: true,
```
kavanagh: Is it safe to make it on by default yet? Yes we should probably have it default to on when we're happy with the UI, but are we ok with doing that right now?

browser/js/app-initializer.js
```diff
@@ -146,11 +148,15 @@ export class AppInitializer {
...
+				if (this.enabledFeatures.feedback) {
+					feedback.init();
+				}
+
 				if (this.enabledFeatures.roe) {
 					roe.init(flags);
 				}
 
-				if (this.enabledFeatures.evenMoreJanky) {
+				if (this.enabledFeatures.evenMoreJanky && flags.get('qualtrics')) {
```
kavanagh: Is this the correct place for the flag?  I would have though it should be on line 151.

dapenguin: Poop! Good spot

PR #1270, opened 2018-06-14
Pass appInfo into the feedback function

dapenguin:
🐿 v2.8.0

### DIFF
--- a/browser/js/app-initializer.js
+++ b/browser/js/app-initializer.js
@@ -149,7 +149,7 @@ export class AppInitializer {
 				}
 
 				if (this.enabledFeatures.feedback && flags.get('qualtrics')) {
-					feedback.init();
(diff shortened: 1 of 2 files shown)


browser/js/app-initializer.js
```diff
@@ -149,7 +149,7 @@ export class AppInitializer {
 				}
 
 				if (this.enabledFeatures.feedback && flags.get('qualtrics')) {
-					feedback.init();
+					feedback.init(appInfo);
```
leggsimon: does it need `appInfo`?

dapenguin: Yeah, that's how we get the name of the app (e.g. article, stream, etc.).
I could scrape that info myself from the `<html>` tag, but `n-ui` does that already, so I thought it made sense to use it.

PR #1274, opened 2018-06-18
bump n-express

pixelandpage:
🐿 v2.9.0


doramatadora:
`.scss-lint.yml` is very old - we deprecated scss-lint in favour of stylelint. This file should be deleted

PR #1279, opened 2018-06-20
Filter out sourcepoint errors

alexflorisca:
🐿 v2.9.0

### DIFF
--- a/browser/bundles/o-errors/filter-error.js
+++ b/browser/bundles/o-errors/filter-error.js
@@ -7,14 +7,16 @@ module.exports = ({ exception }) => {
 		/'undefined' is|undefined is|is undefined/i
 	];
 	let windowFtError;
+	let sourcepointError;
 	if(exception) {
 		try {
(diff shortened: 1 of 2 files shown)


browser/bundles/o-errors/filter-error.js
```diff
@@ -7,14 +7,16 @@ module.exports = ({ exception }) => {
 		/'undefined' is|undefined is|is undefined/i
 	];
 	let windowFtError;
+	let sourcepointError;
 	if(exception) {
 		try {
 			let errorString = JSON.stringify(exception);
 			windowFtError = errorFilters.every(rx => rx.test(errorString));
+			sourcepointError = /sourcepoint/.test(errorString);
```
liamkeaton: Why can't this go in the `errorFilters` array?

alexflorisca: Because I need to merge the other two values in there together into one regex and also because I suck at regex

PR #1286, opened 2018-06-27
deliver es6 to modern browsers

doramatadora:
# Front-end delivery improvements
### What is this? 
The third of a refactoring trifecta of client-side JS bundling and delivery. This is _likely_ a breaking version.

### Changelog
- adds a webpack config supporting ES6 bundling*
- refactors app shell and components to use ESM import/export syntax
- builds ES5 bundles and ES6 bundles in parallel (deploying all hashed assets for the `n-ui` app shell to S3 as usual)
- bootstraps with ES6 automatically for modern browsers, using a `<script type="module">`/`<script nomodule>` mustard cutting mechanism

*_but not ESM (ES modules)._ We have two options here: refactor our client-side code as ESM (**we should do this anyway**), with judicious use of import/export, or wait until Babel 7 comes out of beta and use [`babel-esm-plugin`](https://www.npmjs.com/package/babel-esm-plugin) to "transpile" CommonJs modules to ESM. 

**Even without ESM, the bundle savings are significant**. Native ESM does mean [faster JS loading times in the browser](https://hackernoon.com/native-es-modules-ready-for-prime-time-87c64d294d3c); refactoring client-side code is therefore recommended.

### A comparison - inspect using Chrome DevTools' Network tab

- ES5: https://ft-next-test-artefacts.s3-eu-west-1.amazonaws.com/n-ui/test-page/6863/test-page.html
- ES6: https://ft-next-test-artefacts.s3-eu-west-1.amazonaws.com/n-ui/test-page/6899/test-page.html

ES6 is a 61KB saving on the app shell (compare `es5.js` to `vendor.js.es6.js`) even without ESM.

### Release cycle

- [ ] Release beta version of this
- [ ] Test beta version with client-facing apps (some code changes may be required if an app's client-side code mixes `import/export` with default CommonJS exports – the logs are verbose and should help identify such instances; switching to [named exports](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) instead of `module.exports = { thingToImport }` is usually the solution)

## TODO

- [ ] default Resource Hints to ES6 bundles (thanks @wheresrhys)


demo/app.js
```diff
@@ -98,20 +98,24 @@ app.listen(5005)
...
+						'public/main.js.es6.js'
+					];
+					[
+						'o-errors',
+						'vendor',
+						'font-loader'
+					].forEach(bundle => {
+						files.push(
+							`public/n-ui/${bundle}.js`,
+							`public/n-ui/${bundle}.js.es6.js`
```
leggsimon: should this be `public/n-ui/${bundle}.es6.js` instead of having 2 `.js`s

build/webpack/webpack.common.config.js
```diff
@@ -5,9 +5,36 @@ const config = {
...
 };
 
-module.exports = (configKeys = []) =>
+const webpackConfig = (configKeys = []) =>
 	webpackMerge(...configKeys.map(key => config[key]));
+
+function webpackConfigFormula ({
+	includeExternals = false,
+	includeAppShell = false,
+	jsLoader = 'es5'
```
leggsimon: I ❤️ this! Using config objects with defaults!

chee:
@leggsimon the decision is happening here:
https://github.com/Financial-Times/n-ui/pull/1286/files#diff-3384758e6cfb6b011dd41ac0ee26d048R29
(and here https://github.com/Financial-Times/n-ui/pull/1286/files#diff-3384758e6cfb6b011dd41ac0ee26d048R40 )

where browsers that support `type=module` will get the es6 module and ones that don't will get the `nomodule`. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Module_fallback

it works for everyone other than Safari 10.1 i think, which supported module but didn't support the `nomodule` attribute.

build/webpack/webpack.common.config.js
chee: me too

demo/app.js
chee: it looks like they're being built as `${name}.es6.js`, which i think would mean this is right ?

chee:
(there's [a trick](https://gist.github.com/samthor/64b114e4a4f539915a95b91ffd340acc) you can do to have it download but not execute the second script)

edds:
This PR is being closed due to inactivity as part of a Pull Request Bankruptcy. If you feel it’s still useful or relevant please feel free to re-open it.

PR #1295, opened 2018-07-12
Removes marking promo banner

CourtneyLO:


components/n-ui/header/index.js
```diff
@@ -1,9 +1,7 @@
 const OHeader = require('o-header');
-const promoHandler = require('./js/promoHandler');
 const TopicSearch = require('n-topic-search');
 
-function init (flags) {
```
liamkeaton: What's calling this with `flags` as you can remove that from the call now

CourtneyLO: Thanks, can't believe I missed that :)

PR #1297, opened 2018-07-17
Adds .jsx support in build

maxcbc:
- adds jsx support in build (was already there apart from the filetype)


maxcbc:
Always forget about the `?`

maxcbc:
Have made that change

PR #1298, opened 2018-07-18
Update mainFields in webpack

maxcbc:
# Update mainFields in webpack common config
## What
- Update `mainFields` in webpack common config to prefer `browser` over `main`

## Why
- If a file has already been made for the browser, we should use that rather than doing the work again
- Without this we cannot use x-engine with n-ui

### DIFF
--- a/build/webpack/common.js
+++ b/build/webpack/common.js
@@ -46,7 +46,7 @@ module.exports = {
 
 		// package.json / bower.json
 		// fields for package resolution
-		mainFields: ['main', 'browser'],
+		mainFields: ['browser', 'main'],
 
 		// package.json / bower.json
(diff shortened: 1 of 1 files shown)


maxcbc:
@doramatadora I had a talk with @i-like-robots yesterday and he was of the opinion that theres a fair few dependencies around using the 'module' mainFile that are using gammy non-spec compliant implementations of ES modules so we might want to avoid this. I'm no expert on this stuff myself though.

doramatadora:
@maxcbc it won't make a difference at this stage because we're still babelifying it all. happy to leave it out for now

maxcbc:
Thanks @doramatadora

PR #1306, opened 2018-07-30
Split o ads bundle

alexflorisca:
Putting o-ads initialisation code into a separate bundle to see if it speeds up ads without affecting main site code


adgad:
Looks okay to me. Do apps have to do anything for it to work? (like add the file to gitignore or whatever?)

PR #1310, opened 2018-08-02
Replace Googles document.write with async version

GlynnPhillips:
This used to add Google customer surveys on article pages. This is a
newer version of the code provided by Google so we don't have to do
horrible `document.write`'s

These surveys are currently impossible to force onto the page so
changing this is very much fire and forget. Yay.

 🐿 v2.10.2


GlynnPhillips:
Corresponding PR in next-article https://github.com/Financial-Times/next-article/pull/2905

PR #1322, opened 2018-08-29
Fix call to `nht deploy-hashed-assets` in `publish` job

simonplend:
We need to run `make build-dist` in the `build` job to generate the `assets-hashes.json` file required by
`nht deploy-hashed-assets` in the `publish` job. This is because the `test` job, where `make test` previously triggered `make build-dist`, does not persist any changes in the working directory to the workspace that is shared between jobs - `build` is the only job that does persist changes to the workspace.

 🐿 v2.10.2


adgad:
Makes sense, it's weird that it was in the `test` task in the first place 🤔

PR #1334, opened 2018-10-08
[DO NOT MERGE] Remove haikro from CLI command

adgad:
Removing this for testing out with next-heroku-playground

(incidentally NO idea why n-ui build needs to call haikro build. They should be independent.)


adgad:
We'd have to update 22 repos if we did this 😬 
https://github.com/search?p=2&q=org%3AFinancial-Times+%22nui+build+--production%22&type=Code&utf8=%E2%9C%93

PR #1338, opened 2018-10-17
Add option for no haikro to Cli command

kitkat119:
Checks that the env var NO_HAIKRO does not exist in order to run Haikro build. This is so we can build without Haikro while testing out the new build and deployment process in next-heroku-playground and next-search-page.

@adgad is this ok?


taktran:
For reference, it's for this: https://github.com/Financial-Times/next/issues/212

PR #1354, opened 2018-11-23
Exclude fragment part of URL when searching query params

alexandercurtis:
🐿 v2.11.0


alexandercurtis:
Of course, the location.search already excludes the fragment part!

PR #1356, opened 2018-12-04
Move nht rebuild into n-ui

kitkat119:
🐿 v2.11.0

As part of the etg-heroku project we have been simplifying n-heroku-tools and since `nht rebuild` is only used by n-ui, we felt this would be an appropriate new home for it. 

I have tested it locally using `./bin/nui.js rebuild next-search-page` which successfully triggered search-page to rebuild. I also tested out `./bin/nui.js rebuild --all --serves user-page` with some parts commented out to avoid actually triggering lots of rebuilds, and this worked too. I tested the `build` and `watch` commands within next-search-page and these were also successful.

I have had to refactor rebuild and some parts of the existing cli in order to get code climate to pass. Let me know if you think I could have done this in a better way :)


build/lib/development-keys-path.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+
+const join = require('path').join;
+module.exports = join(process.cwd(), '.env');
```
adgad: I wonder if this can be moved into the file below? Seems a little lonely here, and it's only used in that one place...

adgad: (Discussed in real life) We think we can actually remove most of this `.env` stuff. It's only required for running `nui rebuild` locally, which isn't very often. And can be achieved by doing `export CIRCLE_REBUILD_KEY=blah; nui rebuild`. So we can put that in the README and remove all the code.

README.md
```diff
@@ -120,6 +120,9 @@ In the app (e.g. next-article):
 - `make build` - !! need to do this for each change in either n-ui or front facing app.
 - `make run`
 
+To test the cli locally, use `../n-ui/./bin/nui.js build` or `../n-ui/./bin/nui.js watch` from within a local copy of a user-facing app. To test the rebuild command first set the `CIRCLECI_REBUILD_KEY` environment variable using `export CIRCLECI_REBUILD_KEY={key}` (you can find the key in n-ui's continuous-integration folder in Vault). You can then use `./bin/nui.js rebuild next-search-page` within your local copy of n-ui (perhaps comment out the call to `triggerMasterBuild` in the rebuild function when testing the `./bin/nui.js rebuild --all --serves user-page` command so you avoid actually causing all apps to rebuild ;)).
```
adgad:
```suggestion
To test the cli locally, use `../n-ui/bin/nui.js build` or `../n-ui/bin/nui.js watch` from within a local copy of a user-facing app. To test the rebuild command first set the `CIRCLECI_REBUILD_KEY` environment variable using `export CIRCLECI_REBUILD_KEY={key}` (you can find the key in n-ui's continuous-integration folder in Vault). You can then use `./bin/nui.js rebuild next-search-page` within your local copy of n-ui (perhaps comment out the call to `triggerMasterBuild` in the rebuild function when testing the `./bin/nui.js rebuild --all --serves user-page` command so you avoid actually causing all apps to rebuild ;)).
```

The `/./` part in the path is a reference to the current directory - so when you're already specifying the directory it's redundant.

build/app/cli.js
```diff
@@ -50,15 +53,73 @@ const aboutJson = () => {
...
+const getAllAppsToRebuild = async (allApps, registry, servesType) => {
+	let allAppsToRebuild = [];
+
+	const registryData = await fetch(registry).then(fetchres.json);
+	allAppsToRebuild = registryData
+		.filter(serves(servesType))
+		.map(getRepoName)
+		.filter(repo => repo);
+
+	return allAppsToRebuild;
```
i-like-robots: No need to create and assign the  `allAppsToRebuild` variable, you can just return above =]

build/app/cli.js
```diff
@@ -50,15 +53,73 @@ const aboutJson = () => {
...
+const serves = type => app => type ? app.types && app.types.includes(type) : true;
+
+const getAllAppsToRebuild = async (allApps, registry, servesType) => {
+	let allAppsToRebuild = [];
+
+	const registryData = await fetch(registry).then(fetchres.json);
+	allAppsToRebuild = registryData
+		.filter(serves(servesType))
+		.map(getRepoName)
+		.filter(repo => repo);
```
i-like-robots: `.filter(Boolean)`?

build/app/cli.js
```diff
@@ -50,15 +53,73 @@ const aboutJson = () => {
...
+
+	return allAppsToRebuild;
+};
+
+async function rebuild (options) {
+	const { apps, allApps, servesType, registry = DEFAULT_REGISTRY_URI } = options;
+	let appsToRebuild = [];
+
+	const areAppsToRebuild = (apps.length) || allApps;
```
i-like-robots: No need to wrap the expression in parenthesis here

PR #1359, opened 2019-01-08
HEROKU_SLUG_COMMIT isn't updated by postbuild, so use SOURCE_VERSION

quarterto:


build/app/cli.js
```diff
@@ -28,10 +28,10 @@ const nUiVersion = require('../../package.json').version;
 let appPackageJson;
 
 const getAppVersion = () => {
-	if(process.env.HEROKU_SLUG_COMMIT) {
+	if(process.env.SOURCE_VERSION) {
 		// For apps that use the heroku-postbuild to build, there is no git repository
 		// so we need to enable Dyno Metadata and get the commit hash from environment variable
```
taktran: Minor thing, but no longer part of dyno metadata. Just [buildpack-api](https://devcenter.heroku.com/articles/buildpack-api)

PR #1361, opened 2019-01-10
Add new floodlight tracking alongside existing implementation

remybach:
🐿 v2.11.0

### DIFF
--- a/components/n-ui/tracking/third-party/floodlight.js
+++ b/components/n-ui/tracking/third-party/floodlight.js
@@ -19,6 +19,9 @@ module.exports = function (flags) {
 	const isSignUpForm = /^\/signup/.test(location.pathname);
 	const isSubscriptionConfirmation = /^\/thank-you/.test(location.pathname);
 	const isTrialConfirmation = (l) ? l.getAttribute('data-signup-is-trial') === 'true' : undefined;
+	const isBarrier = !!document.querySelector('[data-trackable="page:product"]');
+	const isSubscriber = window.FT.flags.subscriberCohort;
+	const isRegistered = !isAnonymous && !isSubscriber;
 
 	const spoor = (spoorId) ? spoorId[1] : '';
 	const ts = Date.now();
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "@financial-times/n-gage": "^2.0.2",
     "@financial-times/n-heroku-tools": "^7.4.0",
     "@financial-times/n-test": "^1.7.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
```
remybach: This was needed to get `n-ui` working with a locally running app.

components/n-ui/tracking/third-party/floodlight.js
```diff
@@ -19,6 +19,9 @@ module.exports = function (flags) {
 	const isSignUpForm = /^\/signup/.test(location.pathname);
 	const isSubscriptionConfirmation = /^\/thank-you/.test(location.pathname);
 	const isTrialConfirmation = (l) ? l.getAttribute('data-signup-is-trial') === 'true' : undefined;
+	const isBarrier = !!document.querySelector('[data-trackable="page:product"]');
+	const isSubscriber = self.FT.flags.subscriberCohort;
```
liamkeaton: Where is self defined?

remybach: https://stackoverflow.com/questions/3638887/whats-the-difference-between-self-and-window 😉

liamkeaton: mindblown

I would say `window` is more well known and readable? `self` has been misused as scoped `this` for far too long.

components/n-ui/tracking/third-party/floodlight.js
```diff
@@ -39,16 +42,33 @@ module.exports = function (flags) {
 		document.body.dispatchEvent(event);
 	};
 
-	if (flags && (flags.get('floodlight') && spoorId)) {
+	const gtagEvent = (sendTo) => {
+		self.gtag('event', 'conversion', {
```
liamkeaton:
```suggestion
		window.gtag('event', 'conversion', {
```

PR #1364, opened 2019-01-17
Support multiple teaser tests

saldixon:
This is intended to replace the 'headline testing' code that currently supports A/B testing of only one homepage teaser at any given time.  Will rip out the headline testing code once its replacement, teaser-testing, is fully operational end-to-end.


components/n-ui/tracking/ft/utils/abTestHelpers.js
```diff
@@ -0,0 +1,17 @@
+
+const getTeaserTestContext = function (doc) {
+
+	const teasersUnderTest = [].slice.call(doc.querySelectorAll('[data-trackable-context-teaser-variant]'));
```
i-like-robots: You can use `Array.from()` here (it should be polyfilled by default for older browsers)

PR #1368, opened 2019-01-28
use polyfill.io v3

leggsimon:
Uses v3 of the polyfill service which introduces updated polyfills and improved cache hits


i-like-robots:
Has this been tested in various browsers? Thinking specifically about IE11 which requires the most polyfilling

leggsimon:
I haven't tested it at all. Jake assures me that it will have exactly the same behaviour (and that under the hood we are actually using v3 already)

JakeChampion:
The change in this pull-request will make the polyfill-service use the latest version of the polyfills, which is likely something you would want to test across browsers. If you wanted to use the same polyfills as v2 you need to add a `version` query parameter set to `3.25.1`.

i-like-robots:
I'll pick this up today for testing so we can get it over the line 🏁

i-like-robots:
Tested in IE11 across two apps (search and video) and it was well

leggsimon:
Thank you very much Matt!! 🙇

PR #1370, opened 2019-02-06
fix: forced to latest version of n-feedback to fix conflict on overlay

umbobabo:
🐿 v2.12.0


umbobabo:
Relate to this fix https://github.com/Financial-Times/n-feedback/pull/41

PR #1371, opened 2019-02-06
Temporarily pin n-feedback to attempt fixing broken builds

carlesandres:
🐿 v2.12.0


carlesandres:
Not needed. This was not solving the problem

PR #1381, opened 2019-02-26
NOT TO BE MERGED YET!! Use o-ads as npm dependency

carlesandres:
🐿 v2.12.0


chee:
this would be a mistake.

`o-ads` depends on `o-viewport` and `o-grid`. Which means that if anyone is using any other origami components that depend on those, or those directly, they would end up with two in their bundle. (the one in bower, and the one that is required by npm)
which version of the sass they get (including now `o-colors` and `o-visual-effets`, two more of `o-ads` dependencies) would depend on the order the bundler looks at `includePaths`. 

**FT.com is categorically unprepared to use Origami via npm.**

carlesandres:
Thanks @chee !

PR #1382, opened 2019-02-27
Add audio teaser view tracking

fenglish:
To apply audio teaser view tracking for next-article, next-stream-page and next-front-page.

This PR need to be merged after o-tracking is released.
https://github.com/Financial-Times/o-tracking/pull/199/

 🐿 v2.12.0

### DIFF
--- a/components/n-ui/tracking/ft/events/audio-teaser-view.js
+++ b/components/n-ui/tracking/ft/events/audio-teaser-view.js
@@ -1,0 +1,22 @@
+const oTracking = require('o-tracking');
+
+const audioTeaserView = {
+	init: () => {
+		const getContentIdFromXTeaser = el => el.getAttribute('data-id');
+		const getContentIdFromNTeaser = el => el.querySelector('.o-teaser__heading a').getAttribute('data-content-id');
+
+		oTracking.view.init({
+			selector: '.o-teaser--audio',
+			getContextData: (el) => {
+				return {
+					contentId: getContentIdFromXTeaser(el) || getContentIdFromNTeaser(el),
+					component: 'teaser',
+					type: 'audio',
+					subtype: 'podcast', // only podcast is audio subtype at 03/2019. Need to change when audio has more subtypes.
+				};
+			}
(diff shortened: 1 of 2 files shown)


components/n-ui/tracking/index.js
```diff
@@ -8,6 +10,6 @@ module.exports = {
 		});
 	},
 	scrollDepthComponents: require('./ft/events/scroll-depth-components'),
-	scrollDepth: require('./ft/events/scroll-depth')
-
+	scrollDepth: require('./ft/events/scroll-depth'),
+	view: oTracking.view
```
benbarnett: Out of interest, why does this need to live in n-ui? Can we not use the method directly  from `o-tracking` in the various apps?

Asking as we're potentially bringing in additional weight here to apps that may not require this functionality.

benbarnett: Although it does actually look like this is happening already... just my lack of knowledge of n-ui 🙈

components/n-ui/tracking/ft/events/audio-teaser-view.js
```diff
@@ -0,0 +1,22 @@
+const oTracking = require('o-tracking');
+
+const audioTeaserView = {
+	init: () => {
+		const getContentIdFromXTeaser = el => el.getAttribute('data-id');
+		const getContentIdFromNTeaser = el => el.querySelector('.o-teaser__heading a').getAttribute('data-content-id');
```
benbarnett: ~As this code lives at a global level, it might be wise add a check that `el.querySelector('.o-teaser__heading a')` exists before running `getAttribute()` on it.~

On second thoughts I think we want this to throw so we know about it going wrong?

fenglish: There is not such a deep thought... 🙈
However, I agree with you. We should keep as it is.  `o-teaser--audio` should have `.o-teaser__heading a`. If it didn't had the element, it meant something went really wrong.

PR #1383, opened 2019-02-28
Send new ad-related metrics to Spoor

carlesandres:
This makes `n-ui` do the following:

1. Record a new performance mark for each one of a set of custom events dispatched by `o-ads`. Specifically: 	`oAds.startInitialisation`, `oAds.moatIVTcomplete`, `oAds.apiRequestsComplete` and `oAds.initialised`.
1. Send a new 'event' to `spoor-api` upon receiving `oAds.adServerLoadSuccess` (which marks the end of the page-level preparation process for loading ads) with all those performance marks in its payload.


components/n-ui/ads/js/page-metrics.js
```diff
@@ -0,0 +1,76 @@
+const inMetricsSample = require('./utils').inMetricsSample;
```
andrewgeorgiou1981: Just querying if we want to call this file "page-metrics", will we share this file with the slot-metrics also? or is this intended to be seperate from slot-level metrics?

carlesandres: When we work on the slot-level metrics, we will have to extract some common methods into a separate file metrics-utils. I'd rather leave it like this for now and refactor/rename when we get to work on the slot metrics.

PR #1389, opened 2019-03-12
Improved the way we sample users who send ads metrics

carlesandres:
The ads team has agreed on sending metrics to spoor for only 10% of all users.

Until now, we were using the spoor-id cookie to allocate users in send/no-send groups. However, spoor ids can take several shapes as there are not strong requirements for them and they currently come in, at least, two different shapes.

Using `Math.random` gives us more control over the granularity of the sample size and makes it easier to reason about it, while still achieving the main goal of this sampling which is not flooding Spoor with too many events unnecessarily.

For every page visit a user is placed consistently in a send/no-send group which should make it easier to compare metrics from different requests to Spoor.

### DIFF
--- a/components/n-ui/ads/js/utils.js
+++ b/components/n-ui/ads/js/utils.js
@@ -130,7 +130,7 @@ const inMetricsSample = (() => {
 	const decideInMetricsSample = () => {
 
 		// We are caching the value since we want to be consistent with the user
-		// allocation throughout one same session
+		// allocation throughout one same page visit
 		if (typeof userSendsMetrics !== 'undefined') {
 			return userSendsMetrics;
 		}
--- a/components/n-ui/ads/js/utils.js
+++ b/components/n-ui/ads/js/utils.js
@@ -122,9 +122,25 @@ function getScreenSize () {
 	return window.innerWidth;
 }
 
-function inMetricsSample() {
-	return /spoor-id=3/.test(document.cookie);
-}
+// Fraction of all users that will actually send ads metrics to Spoor
+const METRICS_SAMPLE_SIZE = 0.1;
+
+const inMetricsSample = (() => {
+	let userSendsMetrics;
+	const decideInMetricsSample = () => {
+
+		// We are caching the value since we want to be consistent with the user
+		// allocation throughout one same session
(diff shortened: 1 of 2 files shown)


components/n-ui/ads/js/utils.js
```diff
@@ -122,9 +122,25 @@ function getScreenSize () {
...
+// Fraction of all users that will actually send ads metrics to Spoor
+const METRICS_SAMPLE_SIZE = 0.1;
+
+const inMetricsSample = (() => {
+	let userSendsMetrics;
+	const decideInMetricsSample = () => {
+
+		// We are caching the value since we want to be consistent with the user
+		// allocation throughout one same session
+		if (typeof userSendsMetrics !== 'undefined') {
```
SamiTriki: Looks good, I'm just curious at how this userSendsMetrics is persisted, it's a bit difficult to track. but if I understand correctly, on subsequent calls you should still access the value if it has been defined for the current session?

carlesandres: Yes. The innermost function (`defineInMetricsSample`) has a reference to the outer variable (`userSendsMetrics`) which creates a closure, (i.e. the variable is never garbage-collected) and its value is only accessible from within that function. Check out [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures#Closure) for a better explanation.

carlesandres: By the way, as Andrew pointed to me, 'session' is not correct here as the value is just persisted in browser memory for the current page visit. 

Navigating to a new page, will re-run the logic from scratch, so the user will be placed again in send/no-send group.

components/n-ui/ads/js/utils.js
```diff
@@ -122,9 +122,25 @@ function getScreenSize () {
...
-}
+// Fraction of all users that will actually send ads metrics to Spoor
+const METRICS_SAMPLE_SIZE = 0.1;
+
+const inMetricsSample = (() => {
+	let userSendsMetrics;
+	const decideInMetricsSample = () => {
+
+		// We are caching the value since we want to be consistent with the user
+		// allocation throughout one same session
```
carlesandres:
```suggestion
		// allocation throughout one same page visit
```

PR #1397, opened 2019-03-26
Upgrade to o-ads v11 and ESMify the ads component

alexflorisca:
🐿 v2.12.2


alexflorisca:
@carlesandres Thanks for looking at it. I've tested across next-article. The process for pulling in n-ui is exactly the same for front-page and stream-page so I feel like it's enough of a test.

Regarding the code climate errors, they are the same errors that have been marked as "Fixed" before, but have flagged up again because I changed from `module.exports` to `export default` so it thinks it's a new function. I'll check if we have a login for this and silence these again

PR #1402, opened 2019-04-15
insert consent tracking in ad first load call

SamiTriki:
🐿 v2.12.3


SamiTriki:
Closing as metrics capture method have been updated

PR #1403, opened 2019-04-17
New ad slot metrics

carlesandres:
This PR upgrades `n-ui` to `o-ads` v12 while achieving two goals:

1) It leverages the new metrics capabilities in `o-Ads` to simplify some logic.

2) It configures new ad-slot-rendering metrics that weren't being tracked before.


components/n-ui/ads/index.js
```diff
@@ -66,20 +46,6 @@ function onAdsComplete (flags, event) {
 
 		if (detail.slot.gpt && detail.slot.gpt.isEmpty === false) {
 			utils.log.info('Ad loaded in slot', event);
-			if (slotsRendered === 0) {
```
alexflorisca: Where has all this code moved to? It looks like the `onAdsComplete` function is now just logging things...should we remove it? Or add the logging to wherever this logic has moved to?

components/n-ui/ads/js/ads-metrics.js
```diff
@@ -0,0 +1,80 @@
...
+];
+
+function sendMetrics (eventPayload) {
+	if (inAdsMetricsSample()) {
+		nUIFoundations.broadcast('oTracking.event', eventPayload);
+	}
+}
+
+function setupAdsMetrics () {
+	oAdsUtils.setupMetrics(metricsDefinitions, sendMetrics);
```
alexflorisca: To me it's a little confusing passing `sendMetrics` as a callback. In `o-ads.utils` there is also a function called sendMetrics, which calls this `callback` on the last line. A neater solution could be passing the `inAdsMetricsSample` as a parameter here, delete the `sendMetrics` function above and use the `sendMetrics` function in o-ads to do the broadcasting of the event.

components/n-ui/ads/index.js
carlesandres: We are not doing anything special with the 'first' slot rendered now. We treat like any other slot.
I noticed this was only logging but it was not part of my work to remove this logging functionality. 
I am happy to remove it as part of a separate PR. @andrewgeorgiou1981, @SamiTriki @alevito what are your thoughts?

components/n-ui/ads/js/ads-metrics.js
carlesandres: The main reason for having `sendMetrics` as a callback is because we don't want users of `o-ads` to be force to use Spoor API for their metrics. This approach enables greater flexibility. 

This not only keeps `o-ads` unaware of `Spoor API` but it also prevents it from needing `n-ui-foundations` as a dependency.

Adding the metrics sampling to `o-ads` is already on my radar. I plan to add it as an optional config parameter on the metrics config objects. I just didn't want to overdo this PR, specially since we want to start collecting metrics asap.

components/n-ui/ads/index.js
alexflorisca: Makes sense. I'll add a card on trello for this

components/n-ui/ads/js/ads-metrics.js
alexflorisca: Ah ok cool makes sense. In that case, could we rename the `sendMetrics` and `callback` variables in the o-ads.utils? I'll message you about this separately as it's not relevant to this PR. Good work!

PR #1426, opened 2019-05-13
Wip permutive trial

andrewgeorgiou1981:


browser/layout/wrapper.html
```diff
@@ -135,6 +135,17 @@
...
 		{{/if}}
+		{{#if @root.flags.AdsPermutive}}
+			<div>
+				<div class="o-permutive" data-o-component="o-permutive"
+				data-o-permutive-auoto-init="false"
+				data-o-permutive-id="e1c3fd73-dd41-4abd-b80b-4278d52bf7aa"
+				data-o-permutive-key="b2b3b748-e1f6-4bd5-b2f2-26debc8075a"
+				data-o-permutive-page-type="article"
+				data-o-permutive-apiendpoint-user="https://ads-api.ft.com/v1/user"
+				data-o-permutive-apiendpoint-content="https://ads-api.ft.com/v1/content/5cfae92e-6cc5-11e9-80c7-60ee53e6681d"></div>
```
SamiTriki: We're not going to hardcode the content uuid value right? Where can we pass it from?

andrewgeorgiou1981: i've removed this 'declarative' config in favour of passing config to the init function. This way we can get the content UUID in the same way as is done for o-ads.

PR #1429, opened 2019-05-14
Ads - including Permutive origami component for a trial - all functio…

andrewgeorgiou1981:
Ads - including Permutive origami component for a trial - all functionality is behind a flag and will not be enabled for real users.

[Permutive](https://developer.permutive.com/) is a Data Management Platform and is being trialed as a replacement for SalesForce DMP (fka Krux). 
The request to run this trial has come from the commercial and ad-operations team.

Permutive would be used for behavioural profiling of users on a website for the purpose of segmenting them for targeting advertising.
This PR is designed to trial the Permutive functionality behind a flag on FT.com; the intention is to enable the functionality only for FT and Permutive devs to ensure that data is being passed end-to-end in the correct way. 
A new origami component has been created, [o-permutive](https://registry.origami.ft.com/components/o-permutive@1.0.0-beta.2/readme?brand=master)  which wraps permutive's client-side code and integrates it with the FT ad-stack. Further details can be found in the origami registry readme.

### DIFF
--- a/components/n-ui/ads/index.js
+++ b/components/n-ui/ads/index.js
@@ -62,7 +62,7 @@ function onAdsComplete (flags, event) {
 
 export default {
 	//Permutive trial - config
-	let oPermConf = {
+	const oPermConf = {
 		'appInfo' : {
 			'appName' : 'article',
 			'contentId' : '5cfae92e-6cc5-11e9-80c7-60ee53e6681d'
--- a/components/n-ui/ads/index.js
+++ b/components/n-ui/ads/index.js
@@ -61,6 +61,21 @@ function onAdsComplete (flags, event) {
 }
 
 export default {
+	//Permutive trial - config
+	let oPermConf = {
+		'appInfo' : {
+			'appName' : 'article',
+			'contentId' : '5cfae92e-6cc5-11e9-80c7-60ee53e6681d'
+		},
+		'publicApiKeys' : {
+			'id' : 'e1c3fd73-dd41-4abd-b80b-4278d52bf7aa',
+			'key' : 'b2b3b748-e1f6-4bd5-b2f2-26debc8075a3'
+		},
+		'adsApi' : {
+			'user' : 'https://ads-api.ft.com/v1/user',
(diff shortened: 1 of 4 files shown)


andrewgeorgiou1981:
> @andrewgeorgiou1981 It all looks good to me. My only concern would be if any of the keys exposed in the code need to be kept private but I imagine that, since you've committed them here, they don't.

Thanks Carles, the keys are public keys, i've tried to make that clear in the code. However when this goes out of trial stage i'll see about making them env vars. Though they are public api keys, which i've added to secret squirrel for now

components/n-ui/ads/index.js
```diff
@@ -59,13 +60,28 @@ function onAdsComplete (flags, event) {
...
+const oPermConf = {
+	'appInfo' : {
+		'appName' : 'article',
+		'contentId' : '5cfae92e-6cc5-11e9-80c7-60ee53e6681d'
+	},
+	'publicApiKeys' : {
+		'id' : 'e1c3fd73-dd41-4abd-b80b-4278d52bf7aa',
+		'key' : 'b2b3b748-e1f6-4bd5-b2f2-26debc8075a3'
+	},
+	'adsApi' : {
```
alexflorisca: Are you not missing the /context endpoint for stream pages?

andrewgeorgiou1981: yes, however this is intentional at this stage. The trial is initially just checking that data gets passed through end-to-end to Permutive in the correct way. So just the metadata is just for articles at this stage.

alexflorisca: Cool cool cool 👍

i-like-robots:
Could you update the PR description to include some info about what Permutive is and who asked for it/is ultimately responsible for it?

andrewgeorgiou1981:
> Could you update the PR description to include some info about what Permutive is and who asked for it/is ultimately responsible for it?

I've updated the PR with a description of why/how this trial has been set up and who made the request. Lmk if i should add more details.

PR #1437, opened 2019-05-28
Add Snyk to components

jenniferemshepherd:
We are installing [Snyk](https://snyk.io) in all of our repos for security reasons. It will patch or ignore known vulnerabilities. Currently Snyk is enabled on our repos (config and results can be viewed from the Snyk webpage) but we want to install it on each project for increased security when building. See https://github.com/Financial-Times/next/issues/365


jenniferemshepherd:
closing due to duplication with #1461

PR #1440, opened 2019-05-31
Remove permutive to get builds passing again

benbarnett:
I know this is a little out there but I can't seem to figure out how to get n-ui builds passing again when this package is rendered.

Not necessarily proposing that we merge this, but the PR demonstrates the issue.

@andrewgeorgiou1981 @alexflorisca would you mind taking a look at this?


andrewgeorgiou1981:
I'm taking a look into this. Seems that the o-perutive component is causing the build to fail. 
Its fine to merge this PR from my side - we were just testing out the concept with Pemutive, we can merge this PR to remove permutive for now while investigating the issue.

PR #1441, opened 2019-06-03
Receive barrier tracking properties in a single object rather than in multiple variables

jkerr321:
This is the first step to moving tracking for barriers out of `n-ui`. This moves the computation of tracking values into next-product and sends them as one object to `n-ui` for attaching to an `o-tracking` event. 

We are doing this work now as enhancements to barrier tracking have been requested by the data analytics team and rather than 'pollute' `n-ui` with more custom barrier tracking we are moving it out as much as possible to `next-product`.

Additional work is required at a later stage to send this event directly from `next-product` but is not being done right now as it involves updating tracking code for the standalone barrier in the `epaper` repo which is also using the `n-ui` barrier tracking code at the moment.

Once that additional work has been done, the associated file in `n-tracking` will not be needed (FYI @ifeanyiisitor and @i-like-robots, we will keep you up to date with that) - https://github.com/Financial-Times/n-tracking/blob/b848cdfb1dacf3be841dfc0407ab8154c19af9d6/src/trackers/barrier-view-tracking.js

The PR for the work in `next-product` is here - https://github.com/Financial-Times/next-product/pull/742. The `next-product` PR must be merged before the `n-ui` PR.

The associated Trello card for this work is here - https://trello.com/c/YLZER8cN

### DIFF
--- a/components/n-ui/tracking/ft/index.js
+++ b/components/n-ui/tracking/ft/index.js
@@ -5,10 +5,6 @@ const nextEvents = require('./next-events');
 const abTestHelpers = require('./utils/abTestHelpers');
 const broadcast = require('n-ui-foundations').broadcast;
 
-function nodesToArray (nodelist) {
-	return [].slice.call(nodelist);
-}
-
 function getRootData (name) {
 	return document.documentElement.getAttribute(`data-${name}`);
 }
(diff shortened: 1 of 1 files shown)


i-like-robots:
This will make moving to n-tracking much much easier for us all, thank you!

PR #1446, opened 2019-06-07
Remove barrier specific tracking

remybach:
https://trello.com/c/Hme2a2xv/1286-5-move-barrier-tracking-from-n-ui-to-next-product

🐿 v2.12.3

### DIFF
--- a/components/n-ui/tracking/ft/index.js
+++ b/components/n-ui/tracking/ft/index.js
@@ -126,27 +126,15 @@ const oTrackingWrapper = {
 				pageViewConf.context['teaser_tests'] = abTestHelpers.getTeaserTestContext(document);
 			}
 
-			// barriers
-			let barrierType = document.querySelector('[data-barrier]');
-
-			if (barrierType) {
-				pageViewConf.context.barrier = true;
-				pageViewConf.context.barrierType = barrierType.getAttribute('data-barrier');
-			};
-
 			// FIXME - should not fire on barriers, but needs to be around for a while data analytics fix their SQL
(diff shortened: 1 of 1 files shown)


components/n-ui/tracking/ft/index.js
```diff
@@ -126,27 +126,14 @@ const oTrackingWrapper = {
...
 
-			if (barrierType) {
-				const customTracking = JSON.parse(barrierType.getAttribute('data-tracking-object'));
-
-				broadcast('oTracking.event', Object.assign({
-					category: 'barrier',
-					action: 'view'
-				}, context, customTracking));
+			if (window.FT && window.FT.nUi) {
+				window.FT.nUi.trackingContext = context;
```
i-like-robots: I'm guessing this is being set so this data can be picked up from the outside... what will pick it up?

remybach: https://github.com/Financial-Times/next-product/pull/752/files#diff-096103d8490b885f9d5037baea6aea80R82 and https://github.com/Financial-Times/next-epaper/pull/35/files#diff-a93ac81418745fbaa6d1e0e4a4171132R20

i-like-robots: Could you add a comment to the code with these references? It's scary setting a global and not knowing who or what consumes it!

remybach: I've added a more generic comment... direct references tend to get outdated and are unlikely to get updated when this gets used elsewhere. At least there's some indication and a "why" now... hopefully that will suffice?

i-like-robots: Thanks!

PR #1464, opened 2019-07-03
Update o-ads to v12.8

carlesandres:
This brings a few small changes to ads metrics.
 
🐿 v2.12.4


carlesandres:
Closed temporarily as there seems to be a problem

PR #1472, opened 2019-07-29
feat: notify the Change API on deployments

sjparkinson:
This repo has a custom deployment target defined in it's `Makefile`, so we must explicitly call the `change-api` target that is defined in n-gage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ deploy-s3:
 	# deploy to hashed urls on s3
 	nht deploy-hashed-assets --directory public/n-ui --monitor-assets
 
-deploy: deploy-s3 npm-publish rebuild-user-facing-apps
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -133,6 +133,7 @@ deploy-s3:
 	nht deploy-hashed-assets --directory public/n-ui --monitor-assets
 
 deploy: deploy-s3 npm-publish rebuild-user-facing-apps
+	@$(MAKE) change-api
```
quarterto: is there a particular reason to do it this way instead of as another dependency of `deploy` (i.e. after `rebuild-user-facing-apps`?)

sjparkinson: None! Hadn't thought about using it as a target dependency. Will update it to that.

PR #1482, opened 2019-08-20
ADSDEV-185 Ad unit shows as unclassified

alevito:
This should fix the bug introduced by the removal of Krux in https://github.com/Financial-Times/n-ui/pull/1480

The main changes here, excepted for reverting the previous temporary fix, are these https://github.com/Financial-Times/n-ui/pull/1482/commits/bdcb98d08682aa27883fbc42573b2644b45ab09d

The `user` and `content` targeting we're not set properly. Subsequently `adUnit` could not be set as was not found in the expected data point returned by `oAds.targeting.get()`.


alevito:
> Good catch! Maybe using `bind` in `addToTargeting` could have solved the problem too, but I like this approach more

In fact the method `addToTargeting` was bound to `ads.api` - maybe this diff is more clear https://github.com/Financial-Times/n-ui/pull/1482/commits/bdcb98d08682aa27883fbc42573b2644b45ab09d

The problem here was that the data passed to `addToTargeting` was being added to oAds targeting under the data point `something`, instead of the correct one - either `user` or `content`.

alevito:
Fixes ADSDEV-185

PR #1483, opened 2019-08-20
 Update footer section title markup for a11y

tbergmen:
DAC_Headings_Issue1 suggests that footer section titles should be h2.

### DIFF
--- a/components/n-ui/footer/partials/menu/template.html
+++ b/components/n-ui/footer/partials/menu/template.html
@@ -3,9 +3,9 @@
 		{{#each @root.navigation.menus.footer.items}}
 		{{#if submenu}}
 		<div class="o-footer__matrix-group o-footer__matrix-group--{{submenu.items.length}}">
-			<div class="o-footer__matrix-title" {{#ifEquals submenu.items.length 2}}aria-controls="o-footer-section-{{@index}}"{{/ifEquals}}>
+			<h2 class="o-footer__matrix-title" {{#ifEquals submenu.items.length 2}}aria-controls="o-footer-section-{{@index}}"{{/ifEquals}}>
 			{{{label}}}
-		</div>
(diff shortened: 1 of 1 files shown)


i-like-robots:
🤔 I disagree... they're not the second most important thing on any page.

adambraimbridge:
Hm, tricky. Do we have anything that explains why `DAC_Headings_Issue1` suggests `H2`s?

tbergmen:
> Hm, tricky. Do we have anything that explains why `DAC_Headings_Issue1` suggests `H2`s?
> 
> ... what's the url for `DAC_Headings_Issue1`?

The issue is documented here:
https://trello-attachments.s3.amazonaws.com/5aaa39b3d9fed818bd80c809/5aabd4eb3100714dabbeaabc/c61472793b3ddb07899a166b986c7fb8/Financial_Times_-_8_August_2019_-_Accessibility_Audit_Retest.pdf#page=25

I think the original issue was solved and they found this footer issue in the re-test, which is on the next page.

adambraimbridge:
Interesting workaround. 
So the new `<h2>` is visually hidden, but it's there to justify using `<h3>` for the headings?

tbergmen:
> Interesting workaround.
> So the new `<h2>` is visually hidden, but it's there to justify using `<h3>` for the headings?

Yes. It is not ideal, but looks like it will work.

dapenguin:
> Interesting workaround.
> So the new `<h2>` is visually hidden, but it's there to justify using `<h3>` for the headings?

Kind of. It also helps separate the footer content from the rest of the page content by providing it with its own heading.

Admittedly it's not 100% ideal, as we should really be giving sighted users and those relying on assistive technology the same content. Would need some input from UX if we wanted the `<h2>` to be visible.

PR #1486, opened 2019-08-27
A11y: Fix subscribe link styled to look like a button to behave like one

remybach:
This was confusing for users of voice activation software as it looks like a button but behaved like a link.

[Trello ticket](https://trello.com/c/RvFGcUpz/143-dacelementroleissue1)

🐿 v2.12.4

### DIFF
--- a/components/n-ui/header/partials/header/sticky.html
+++ b/components/n-ui/header/partials/header/sticky.html
@@ -38,16 +38,19 @@
 							<ul class="o-header__nav-list o-header__nav-list--right" data-trackable="user-nav">
 								{{#outputBlock 'mobile-sticky-header'}}{{/outputBlock}}
 								{{#each @root.navigation.menus.navbar-right-anon.items}}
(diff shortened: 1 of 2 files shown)


components/n-ui/header/partials/header/sticky.html
```diff
@@ -43,6 +43,7 @@
 											class="o-header__nav-{{#ifEquals label 'Subscribe'}}button{{else}}link{{/ifEquals}}"
```
liamkeaton: Could we split this into two <li>'s now? There are now three places we check `{{#ifEquals label 'Subscribe'}}`

i-like-robots:
Could we also ensure that this fix is ported to Page Kit ([here](https://github.com/Financial-Times/dotcom-page-kit/tree/master/packages/dotcom-ui-header))?

remybach:
@i-like-robots Sure. Will take a look 👍

PR #1487, opened 2019-08-28
ADSDEV-189 - Add Google Tag Manager behind a flag

carlesandres:
This adds the Google Tag Manager to the site behind a new flag. 

I have added it to both the `wrapper` and the `vanilla` templates, although only the `wrapper` template seems to be in use at the moment.

It's worth mentioning that this PR is not trying to address the following:

1) Any script that is currently being added to the site via n-ui code will be included twice if added via GTM.
2) Changes made via GTM don't trigger any CI workflow that can prevent us from going live with something breaks our builds*. 

*I have [already experienced](https://circleci.com/gh/Financial-Times/n-ui/9234) how a malformed tag in GTM can break a n-ui build and this wouldn't be immediately noticed by any developer.

### DIFF
--- a/browser/layout/partials/gtm-body.html
+++ b/browser/layout/partials/gtm-body.html
@@ -1,3 +1,1 @@
-<!-- Google Tag Manager (noscript) -->
 <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-NWQJW68" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
-<!-- End Google Tag Manager (noscript) -->
--- a/browser/layout/partials/gtm-body.html
+++ b/browser/layout/partials/gtm-body.html
@@ -1,1 +1,5 @@
+{{#if @root.flags.enableGTM}}
+<!-- Google Tag Manager (noscript) -->
 <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-NWQJW68" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+<!-- End Google Tag Manager (noscript) -->
+{{/if}}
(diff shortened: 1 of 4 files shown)


i-like-robots:
Are you OK to port this over to https://github.com/Financial-Times/dotcom-page-kit/tree/master/packages/dotcom-ui-shell/ as well? I'm happy to help =]

i-like-robots:
> Changes made via GTM don't trigger any CI workflow that can prevent us from going live with something breaks our builds

😱

carlesandres:
> Are you OK to port this over to https://github.com/Financial-Times/dotcom-page-kit/tree/master/packages/dotcom-ui-shell/ as well? I'm happy to help =]

Yup. Once we are happy with the current implementation, I was planning on asking you how to port it to the new architecture

andrewgeorgiou1981:
My understanding is that there is a commercial requirement to add this tag for revenue reasons by mid-September.
I think there are more issues that should be addressed here before we port this over to Pagekit.
I understand that we have new, more rigorous processes for adding third-party tags - both in terms of data-governance and tech reviews, which is a really good step.
However the points that Carles raised around tag changes triggering builds and also the way we ensure that any changes are tracked in version control should be investigated before we port this functionality to pagekit.

edds:
> However the points that Carles raised around tag changes triggering builds and also the way we ensure that any changes are tracked in version control should be investigated before we port this functionality to pagekit.

GTM has versioning built in. Equally the access controls are locked down such that only a very small group of people can publish changes to the tag. We should come up with a process for publishing changes to the tag and document it in the [next wiki](https://github.com/Financial-Times/next/wiki). It can include a list of builds or automated jobs to kick off.

PR #1488, opened 2019-08-30
Move skip links and add block for bottom message slot for: DAC_Popup_focus_Issue1

remybach:
## Feature Description 

> A popup layer is visible when the page starts giving information to users about an offer but is not navigable by Screen reader users or keyboard users.

Since this blocks the content partially for normal users, it was agreed that we should give these messages a higher priority order when tabbing and/or using a screen reader.

[Conversion Trello card](https://trello.com/c/HfW0q71w/1456-dacpopupfocusissue1)
[DAC Trello card](https://trello.com/c/VF4MhvEh/146-dacpopupfocusissue1)

## Screengrabs

Gifs of me tabbing through from the top of the page:

<details><summary>Cookie Consent Banner</summary>

</details>
<details><summary>Marketing Popup Prompt Banner</summary>

</details>

🐿 v2.12.4

### DIFF
--- a/browser/layout/wrapper.html
+++ b/browser/layout/wrapper.html
@@ -139,6 +139,13 @@
 			</script>
 		{{/if}}
 
+		{{!-- It's important for this to be where it is so that there's a good tab order from an a11y perspective. --}}
+		<a data-trackable="a11y-survey-screen-reader" class="n-skip-link" href="https://www.ft.com/accessibility" tabindex="1">Accessibility help</a>
+		<a data-trackable="a11y-skip-to-navigation" class="n-skip-link" href="#site-navigation" tabindex="1">Skip to navigation</a>
+		<a data-trackable="a11y-skip-to-content" class="n-skip-link" href="#site-content" tabindex="1">Skip to content</a>
+		<a data-trackable="a11y-skip-to-footer" class="n-skip-link" href="#site-footer" tabindex="1">Skip to footer</a>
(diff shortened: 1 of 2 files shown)


browser/layout/wrapper.html
```diff
@@ -139,6 +139,13 @@
 			</script>
 		{{/if}}
 
+		{{!-- It's important for this to be where it is so that there's a good tab order from an a11y perspective. --}}
+		<a data-trackable="a11y-survey-screen-reader" class="n-skip-link" href="https://www.ft.com/accessibility" tabindex="1">Accessibility help</a>
+		<a data-trackable="a11y-skip-to-navigation" class="n-skip-link" href="#site-navigation" tabindex="1">Skip to navigation</a>
```
liamkeaton: Don't some layouts not have a navigation?

remybach: That's a good point... although this is how it was in the header file too ¯\\\_(ツ)\_/¯

liamkeaton: But would the header only be included if the layout needed it?

remybach: Nope, always seems to get included: https://github.com/Financial-Times/n-ui/blob/master/browser/layout/wrapper.html#L145

liamkeaton: Cool then this wont be a regression

PR #1491, opened 2019-09-04
ADSDEV-185 GPT site zone unclassified in index page

alevito:
This PR adds the possiblity to handle reponses coming either from Ads Api `v1` or `v2`. This solves the issue observed for stream-pages where gpt zone/site is seen as unclassified.


components/n-ui/ads/index.js
```diff
@@ -22,6 +22,36 @@ function addZone (content) {
 	}
 }
 
+const contentIsV2 = content => !(content && content.dfp);
```
carlesandres: It's a personal preference, but this function would be a bit easier to read if it was written like `contentIsV1`. Negations always make it a bit harder to read.

alevito:
Closing this PR as Ads Sev Huddle decided to reelase Ads Api v2 after` PageKit` work as will be cleaner than doing it with `n-ui`.
A reference to the closed PR will be kept in the Jira ticket.

PR #1492, opened 2019-09-09
Upgrade o-ads to v13.1

carlesandres:
Compared to `v12.8`, version `13.1` of `o-ads` does the following:

1) removes support for 'Krux' as it's no longer used
2) and  adds support for new `addClass` functionality which enables `o-ads` to add a new class to the container slot if instructed by the creative wrapper sent from the ad server.  This will enable the new "sticky" ad functionality.


carlesandres:
`o-ads v13` is not production-ready. We'll be releasing a new version `o-ads v12` and upgrade `n-ui` to it soon

PR #1496, opened 2019-09-23
Remove account settings and portfolio from nav bar

binaryberry:
It's in the myFT menu now.

Related: https://github.com/Financial-Times/next-myft-page/pull/2124

Trello: https://trello.com/c/J7m239TK/3782-release-navonmyft-to-all-users


i-like-robots:
Looks good. Can we ensure this change is also reflected in Page Kit?

PR #1499, opened 2019-10-03
Make floodlight work for single page buy flow form

remybach:
I totally forgot that we had a different URL for the single page form (which is currently what the form is 100% live to.


umbobabo:
@i-like-robots does this need to be ported to dotcom package ?

PR #1507, opened 2019-10-18
Use the new Google robots meta tag attributes

bjfletcher:
Google has introduced new robots meta tag attributes, specifically `max-snippet` and `max-image-preview`. These settings will hopefully improve our SEO performance.

See the [Google doc](https://webmasters.googleblog.com/2019/09/more-controls-on-search.html?utm_source=wnc_20103768&utm_medium=gamma&utm_campaign=wnc_20103768&utm_content=msg_110260908&hl=en-GB) for more on these attributes.

Talk to me, Matt Bailey (Product Manager) or James Webb if you have any
questions about this.

Page Kit PR for the future: https://github.com/Financial-Times/dotcom-page-kit/pull/618

### DIFF
--- a/browser/layout/wrapper.html
+++ b/browser/layout/wrapper.html
@@ -82,6 +82,7 @@
 		{{#if @root.flags.googleSiteVerificationMeta}}
 		<meta name="google-site-verification" content="4-t8sFaPvpO5FH_Gnw1dkM28CQepjzo8UjjAkdDflTw" />
 		{{/if}}
+		<meta name="robots" content="max-snippet:200, max-image-preview:large">
 
 		{{#outputBlock 'head'}}{{/outputBlock}}
 


i-like-robots:
I'm not sure what the effect of this will be as apps may already declare a robots tag themselves, e.g. https://github.com/Financial-Times/next-article/blob/62da871e473537ce0860099e31c3553cce1d833f/views/content.html#L5-L7

I don't know if it's valid to have both or what will happen when the values contradict as above.

bjfletcher:
@i-like-robots Google [says](https://webmasters.googleblog.com/2007/03/using-robots-meta-tag.html) it should be ok:

If the page contains multiple meta tags of the same type, we will aggregate the content values. For instance, we will interpret

<META NAME="ROBOTS" CONTENT="NOINDEX">
<META NAME="ROBOTS" CONTENT="NOFOLLOW">

The same way as:

<META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">

If content values conflict, we will use the most restrictive. So, if the page has these meta tags:

<META NAME="ROBOTS" CONTENT="NOINDEX">
<META NAME="ROBOTS" CONTENT="INDEX">

We will obey the NOINDEX value.

PR #1514, opened 2019-11-28
Origami major cascade

i-like-robots:
Updates all Origami and `n-` dependencies to their latest major versions.


browser/bundles/main.js
```diff
@@ -20,7 +20,7 @@ import { AppInitializer } from '../js/app-initializer';
 // returns {flags, allStylesLoaded, appInfo}
 const app = new AppInitializer();
 
-window.FT.nUi = Object.assign({}, {}, app.env, {
+window.FT.nUi = Object.assign({}, app.env, {
```
chee: unclean

components/n-ui/header/main.scss
```diff
@@ -41,6 +42,7 @@
 				width: 52px;
 			}
 
+			// TODO: clean up
```
chee: lol

i-like-robots: #1515

chee:
does this PR also remove krux?

andrewgeorgiou1981:
> does this PR also remove krux?

This PR removes the bulk of krux code - some krux code was still around as it was being re-used by `permutive` which is the replacement for Krux (There was a branch to completely remove all krux code; however this caused various issues in the past so we made a decision to leave n-ui as-is and make sure that pageKit/n-ads is completly clean of all krux code). 
Bumping o-ads has forced us to remove the majority of krux code though as o-ads v13 and above is not compatible with krux functionality. 
I have made a new release of o-ads v14.1.0 which is backwards compatible with how n-ui gets page and user-meta data and passes it to permutive and does not require any of the n-ui functionality previously name-spaced under 'krux'.

PR #1522, opened 2019-12-06
Update babel-plugin-add-module-exports to support .default

i-like-robots:
This change updates the Babel configuration to ensure that all ESM code transpiled to CJS is
output with both `module.exports` and `module.exports.default` for full compatibility with
integrations either using `require()` or `require().default`.

This is required because we still have a mix of CJS code, including the `n-feedback` component which is currently broken for users of n-ui v9.


i-like-robots:
I have no idea how to make the Karma tests pass. We need to use a version of this Babel plugin which supports interoperability _but_ before the author added more spec-compliant checks.

Unfortunately a version like this was never published and it cannot be installed from Github either.

i-like-robots:
After some `console.log()` based development (old Karma running in headless browser...) I noticed that the helper methods being injected by the Rewire plugin were nuking the `module.exports = a; module.exports.default = a;`.

PR #1529, opened 2020-01-09
Disable GTM for pa11y tests

carlesandres:
Adding the floodlight tag via Google Tag Manager makes a couple of Pa11y test fail ([see this](https://circleci.com/gh/Financial-Times/n-ui/9835))
 
The same way [we did for next-subscribe](https://github.com/Financial-Times/next-subscribe/pull/746/files) , we are now instructing ignoring Pa11y to not look at Doubleclick iframes.

Since ☝️ this doesn't seem to sort out the problem completely, I'm also (temporarily) disabling GTM altogether during pa11y tests the same way we already disable ads. This obviously introduces some blind spots in our a11y testing but I believe this is acceptable for now as n-ui is being phased out. However, I plan to discuss with the a11y team asap to come up with a better solution.

### DIFF
--- a/browser/layout/pa11y-config.js
+++ b/browser/layout/pa11y-config.js
@@ -6,7 +6,7 @@ module.exports = {
 	pa11yData: [
 		{
 			rootElement: 'html',
-			hideElements: '.n-skip-link',
+			hideElements: '.n-skip-link, iframe[src*=doubleclick]',
 			page: {
 				headers: {
 					'FT-Flags': 'ads:off,javascript:off'
@@ -15,7 +15,7 @@ module.exports = {
 		},
 		{
 			rootElement: 'html',
-			hideElements: '.n-skip-link',
+			hideElements: '.n-skip-link, iframe[src*=doubleclick]',
 			page: {
 				headers: {
 					'FT-Flags': 'ads:off,javascript:on'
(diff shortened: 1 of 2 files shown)


browser/layout/pa11y-config.js
```diff
@@ -6,19 +6,19 @@ module.exports = {
 	pa11yData: [
 		{
 			rootElement: 'html',
-			hideElements: '.n-skip-link',
+			hideElements: '.n-skip-link, iframe[src*=doubleclick]',
 			page: {
 				headers: {
-					'FT-Flags': 'ads:off,javascript:off'
+					'FT-Flags': 'ads:off,javascript:off,enableGTM:off'
```
i-like-robots: Is this belt and braces? Am I right to think there should be no ad iframe if the ads and GTM flags are off?

carlesandres: It's not. For now, gpt.js is still loaded by o-ads and ads still load when the `enableGTM` flag is off. We might change this in the future, but not yet.

browser/layout/pa11y-config.js
```diff
@@ -6,19 +6,19 @@ module.exports = {
 	pa11yData: [
 		{
 			rootElement: 'html',
-			hideElements: '.n-skip-link',
+			hideElements: '.n-skip-link, iframe[src*=doubleclick]',
```
GlynnPhillips: Why are we targeting the `iframe` and also using the `enableGTM` flag?

carlesandres: Sorted now

PR #1530, opened 2020-01-13
Remove the adsDisableMetricsSampling flag 🐿 v2.12.5

alexflorisca:
Removing the adsDisableMetricsSampling flag and making the feature permanent. No user facing changes should be visible

### DIFF
--- a/components/n-ui/ads/index.js
+++ b/components/n-ui/ads/index.js
@@ -13,7 +13,7 @@ window.oAds = Ads;
 function initOAds (flags, appName, adOptions) {
 	const initObj = oAdsConfig(flags, appName, adOptions);
 
-	setupAdsMetrics(flags && flags.adsDisableMetricsSampling);
+	setupAdsMetrics();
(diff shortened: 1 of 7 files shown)


SamiTriki:
The change looks good, do you think it's worth looking at the setupAdMetrics method since we don't rely on a flag anymore and that the method still had a flag as an argument @carlesandres

components/n-ui/ads/index.js
```diff
@@ -13,7 +13,7 @@ window.oAds = Ads;
 function initOAds (flags, appName, adOptions) {
 	const initObj = oAdsConfig(flags, appName, adOptions);
 
-	setupAdsMetrics(flags && flags.adsDisableMetricsSampling);
+	setupAdsMetrics();
```
carlesandres: I think this might be necessary to ensure we don't change the current behaviour:
```suggestion
	setupAdsMetrics(true);
```

PR #1533, opened 2020-01-16
Add change API so that runbook changes get auto pushed

alicebartlett:
So that the NUI runbook gets pushed to Bizops, we have to hook it up to the change API which we can do with circle ci.

And example of this change elsewhere: https://github.com/Financial-Times/next-feedback-api/blob/7d4504beff1b15c3b648f31d962aab4a788c949b/.circleci/config.yml#L141-L146


alicebartlett:
@pixelandpage would you mind having a look at this little baby pull request for me?

